PR #15, opened 2021-03-11
Lpscrypt/work/figma0.5 topbar feedback

LPSCRYPT:
topbar edits

ran npm update which updates many of the package versions


src/App.tsx
```diff
@@ -86,7 +96,9 @@ function App() {
                   />
                 </h1> */}
                 {/* TODO: move to styling file */}
-                <DomainsGlobal domain={'ROOT'} />
+                <div style={{ position: 'absolute', top: '0' }}>
```
benefacto, line 99: If we're using most of our styling in SCSS classes, can we be consistent about that and not use inline styles?

benefacto (changes requested):
Can you merge develop into your branch and resolve any conflicts? Looks like you're two commits behind and we want branches to be up-to-date with develop before we merge them into it. Once that's done and my other comment is resolved, this should be good for me to approve and squash merge in.

src/App.tsx
LPSCRYPT: ok

benefacto: Will be addressed in another pull request

benefacto (approved):
LGTM!

PR #22, opened 2021-03-14
work/josh/render shop and profile

Joshua-Jack:
I put the shop and profile back into the top bar and took out the claim system from the shop.


src/components/topbar/profile/profile.tsx
```diff
@@ -42,21 +42,23 @@ const Profile: FC = () => {
 
   return (
     <>
-      <img src={elon} alt="" className="profilepic" onClick={showOwner} />
-
+      {owned.isJust() && (
```
benefacto, line 45: Are we sure that we want to use this operator? Or do we only want to render the `img` if `owned.isJust()` is true? If so, we'd probably want something like `owned.isJust()?  : null` instead.

src/App.tsx
```diff
@@ -23,24 +23,6 @@ function getLibrary(provider: any): Web3Provider {
 }
 function App() {
   const context = useWeb3React<Web3Provider>();
```
benefacto, line 25: Do we need this line or can it be removed or commented out for now?
```src/App.tsx
  Line 25:9:  'context' is assigned a value but never used  @typescript-eslint/no-unused-vars```

File: src/components/topbar/shop/shop.tsx [hunk 3]
```diff
@@ -1,10 +1,8 @@
 import { FC, useState, useMemo } from 'react';
```
benefacto, line 1: Can we remove/comment some unused variables in this file?
```
src/components/topbar/shop/shop.tsx
  Line 6:8:  '_' is defined but never used         @typescript-eslint/no-unused-vars
  Line 8:8:  'Outgoing' is defined but never used  @typescript-eslint/no-unused-vars
  Line 9:8:  'Claims' is defined but never used    @typescript-eslint/no-unused-vars
```

benefacto (changes requested):
Would you mind taking a look at some of my comments? If you feel like they should be addressed, feel free to do so. Otherwise, we can handle them later. Either way, let me know when you're ready for this to be merged.

src/App.tsx
Joshua-Jack: On it now

src/components/topbar/shop/shop.tsx
Joshua-Jack: On it now

src/components/topbar/profile/profile.tsx
Joshua-Jack: done

benefacto (approved):

PR #23, opened 2021-03-18
Lpscrypt/work/figma0.5 grid

LPSCRYPT:
grid view for domains


benefacto (approved):
LGTM! The responsiveness is not yet there for mobile/tablet devices yet but the grid view should look reasonable on most desktops.

PR #28, opened 2021-03-20
wallet styles

Joshua-Jack:
I have the initial styles done Louis is going to polish the rest as I work on the global search


benefacto (approved):
LGTM!

PR #30, opened 2021-03-20
Work/josh/wallet styles

Joshua-Jack:
Fix deletion issue


benefacto (approved):
This should fix develop

PR #31, opened 2021-03-21
Work/josh louis/demo

Joshua-Jack:
Demo


benefacto (approved):
This is definitely going to need some refactoring before being production-ready (in particular, removing pixel widths/heights for elements & effects and ensuring component aren't reloading unnecessarily when the timer is operating). But, since there is a need to demo this soon, I'm approving its merge into develop with the hope that refactoring will be done later.

PR #35, opened 2021-03-24
Refactored Files

Joshua-Jack:
I have the file system arranged in this order 

component 
  related css folder 
               img folder &
               test folder 

Theres still some tweaks to be done I want to get these pushed so me and louis can hop on a call an he can filter the non needed pictures and css files that need to be correlated.


benefacto (approved):
This is a major improvement to the front end's folder and file structure. Now, it's much easier to tell which SCSS is related to which component and (eventually) which tests. There's certainly more to do but this is really awesome. Great job!

PR #42, opened 2021-04-20
Mailer updates

Remscar:
Fixed mailing list


benefacto (approved):

benefacto (changes requested):
We probably want the modal to close on submit button click but otherwise this seems to work fine!

PR #44, opened 2021-04-28
Latest fixes

colbr:
Got a bunch of fixes to push MVP over the line


src/components/topbar/wallet/wallet.tsx
```diff
@@ -85,135 +108,6 @@ function getErrorMessage(error: Error) {
   }
 }
 
-function ChainId() {
```
Joshua-Jack, line 88: Why was this removed? This logic is necessary but please explain the reason behind it.

Joshua-Jack (changes requested):
Left some comments once those are cleared I will push

Joshua-Jack:
finished the checks

PR #50, opened 2021-05-31
New architecture merge into default branch

colbr:
We branched off a while ago to work on the new front-end architecture. It's at a stage where we're ready to merge back in.


stacy-hunt (approved):
Looks awesome!

PR #60, opened 2021-06-06
Better support for browsers which don't support the 'backdrop-filter' CSS property

colbr:
For browsers which don't support the `backdrop-filter` CSS property:

- Elements which apply the class `.blur` will have more opaque backgrounds
- Overlay background is more opaque

### DIFF
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -138,20 +138,6 @@
 	--animation-time-long: 0.7s;
 }
 
-@supports not (backdrop-filter: blur(30)) {
-	:root {
-		--color-primary: red !important;
-	}
-
-	.blur {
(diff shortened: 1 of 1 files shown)


colbr:
Browser which doesn't support `backdrop-filter` (Firefox 89.0)
Screen Shot 2021-06-06 at 4 33 34 PM
Screen Shot 2021-06-06 at 4 33 28 PM

Browser which does support `backdrop-filter` (Chrome 91.0.4472.77)
Screen Shot 2021-06-06 at 4 33 43 PM
Screen Shot 2021-06-06 at 4 33 39 PM

Not perfect, but it was previously unusable on Firefox. The more opaque backgrounds will work until we have something more solid in place.

Remscar (approved):

PR #61, opened 2021-06-08
Post-release tweaks

colbr:
- [x] Fade in background image when it's finished loading
- [x] Colored background while page is loading (Previously)

As per:

https://user-images.githubusercontent.com/12437916/121116614-58329100-c806-11eb-80ba-4781e681e23e.mov


stacy-hunt (approved):
Nice touch!

PR #67, opened 2021-07-01
Transfer ownership flow

zchlm:
https://trello.com/c/0dtBFzfy/76-transfer-ownership-flow

As this is my first PR, any feedback would be appreciated. 
I've left a number of TODO's for back end dev to implement.
Transfer flow can be triggered on NFT view, not via the grid view three dots as per Figma design. I can do this if needed.


src/lib/providers/TransferProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+	const submit = async (params: TransferSubmitParams) => {
+		console.log(params);
+		addNotification(`Started transfer`);
+		setTransferring([...transferring, params]);
+
+		const finishTransferring = async () => {
+			setFinishedTransferring(params);
+		};
+
+		// TODO
```
Remscar, line 43: Make sure to indicate what TODO's are please, or what it should be doing instead

src/containers/TransferOwnership/TransferOwnership.tsx
```diff
@@ -0,0 +1,179 @@
...
+			image,
+			domain,
+			creatorId,
+			walletAddress
+		});
+	};
+
+	const confirm = () => submitTransfer();
+
+	// Todo
```
Remscar, line 61: Not sure what this todo is supposed to be doing, and if it's related to the commented out code

zchlm: I've removed this commented out code as it's no longer needed

src/lib/providers/TransferProvider.tsx
zchlm: @Remscar I've cleaned up this. See latest commit

src/lib/providers/TransferProvider.tsx
```diff
@@ -26,36 +35,52 @@ const TransferProvider: React.FC<TransferProviderType> = ({ children }) => {
...
-		// TODO: Back end transfer request.
-		const transfer = async () => {
-			setFinishedTransferring(params);
-		};
+			addNotification(`Started transfer`);
+			setTransferring([...transferring, params]);
 
-		await transfer();
-	};
+			tx.wait()
```
Remscar, line 64: Avoid using `.then` in this code base. Since we can use async/await functions it's best for us to only use async/await since mixing the two can easily cause bugs and confusion later on.

You can imitate this exact same logic by defining an async function here, then calling it, but not awaiting that call. (Take a look at src/providers/MintProvider.tsx and the `finishMinting` function that gets called if you'd like an example)

Remscar:
Ideally there should be a loading/waiting screen while the dApp wait for the user to either accept/cancel the transfer transaction. Currently after I hit the "confirm" button the modal closes and i'm shown the domain page as if nothing happened.

zchlm:
> Ideally there should be a loading/waiting screen while the dApp wait for the user to either accept/cancel the transfer transaction. Currently after I hit the "confirm" button the modal closes and i'm shown the domain page as if nothing happened.

@Remscar That definitely shouldn't happen. It should show a notification when the transfer has started and another when completed. Here's how the flow works for me.

https://user-images.githubusercontent.com/8256446/128590636-7420c5d6-4b88-4983-b9fe-1da91cdf3b17.mov

Only thing I couldn't work out is how to refresh the domain after the transfer has finished.

Remscar:
@zchlm Check this out: https://drive.google.com/file/d/1f1ct4GEDRizfB4kr-MPUiiQL6iigDWc9/view?usp=drivesdk

This should explain what I mean better

zchlm:
@Remscar Thanks for the video, that cleared things up!!

We now have the loading state when the transfer is submitted: 

image

Preview: https://bafybeidhl7p3qnwxxxr6pxlbsbtrpqutbxqmmhc5v2f4lh7pbe3gbmsd3a.on.fleek.co

If there's anything else, let me know

Remscar:
Looks great!

Remscar (approved):

PR #68, opened 2021-07-02
Tweaks from feedback (Unfinished)

colbr:
Just creating an early PR so Stacy can check the built version. **Don't merge yet!**

- [x] Search flickers when you click on it
- [x] Search should have mouseover state
- [x] Search should be single line
- [x] Replace "!" with "." in both search messages
- [x] Contain preview card images
- [x] Add "View Domain" button to preview card
- [x] Should be able to link to NFT view of card
- [x] Button for moving back to table view of domain
- [x] Clicking preview image should open lightbox
- [x] Table elements not vertically centered
- [x] Offers page slightly moves as you click headers - should not happen
- [x] Consolidate offers into a single tab with additional column to simplify UX
- [x] After you have minted an NFT, if you click the notification panel to see the minted NFT, clicking the notification button / zero icon does not close the modal
- [x] Change terminology for ‘enlisting’ to ‘waitlist’ and ‘join waitlist’ respectively
- [x] Not obvious that numerical value is a price to be entered in UX
- [x] Can't see NFTs you own
- [x] Need better error notifications if you don't have enough tokens to mint/stake
- [x] Refreshing page should keep you on current zNA (not move you back to the root)
- [x] Use latest background from Ethan’s design and scale rather than tile it
- [x] USD values should always show two decimals
- [x] Need ellipses or wrap for long content in NFT card
- [x] More text line height in Notifications
- [ ] Back button - remember the last subdomain you went forward to within that domain
- [ ] Make description box by default one line high, and expand as you type
- [ ] Clicking on a fulfilled NFT in offers page should bring you to that actual NFT in zNS
- [ ] When Minting an NFT, offer should fade in background image and text so that it does not abruptly change position
- [ ] Can't upload large images (i.e. > 100 mb) ... how does Makersplace, etc. deal with this?
- [ ] Long load times — can take several minutes to load
- [ ] Remove ugly input scroll boxes on Windows
- [ ] Higher-res background

Unable to complete:
- Need a way to deny offers -- There's nothing on the back-end for this one yet
- Show most recently minted NFT first -- There's no date data on the back-end yet (to my knowledge)


Remscar (approved):
We need to get this in

PR #70, opened 2021-07-13
Leo zAuction midlayer integrations

Remscar:


src/containers/MakeABid/MakeABid.tsx
```diff
@@ -57,9 +63,20 @@ const MakeABid: React.FC<MakeABidProps> = ({ domain, onBid }) => {
...
 	const [error, setError] = useState('');
+	const [wildBalance, setWildBalance] = useState(0);
+
+	//- Web3 Wallet Data
+	const walletContext = useWeb3React<Web3Provider>();
+	const { account, chainId } = walletContext;
+	const zAuctionAddress = addresses[chainIdToNetworkType(chainId)].zAuction;
+	const wildContract: ERC20 = useZnsContracts()!.wildToken;
+
+	const getBalance = wildContract
```
Remscar, line 74: This is not the most performative way of doing this and will cause the dApp to query for an updated token balance on every re-render of this component. The proper way of doing this is by using `React.useEffect`

`React.useEffect` (or `useEffect`) is how you can call async logic such as this, but only when certain conditions happen (such as state variables changing, or once when a component loads)

By replacing this code with something like this:
```
React.useEffect(() => {
		if (!account) {
			return; // better safe than sorry
		}

		const fetchTokenBalance = async () => {
			const balance = await wildContract.balanceOf(account);
			setWildBalance(parseInt(ethers.utils.formatEther(balance), 10));
		}
		fetchTokenBalance();
	}, [wildContract, account]);
```
You make it so that it only fetches the balance when the `account` changes or `wildContract` changes.


Another couple of hints here:
- You declared a variable `getBalance` here which is the promise chain to get the balance. `getBalance` was an un-used variable, this is a sign that you're possibly doing something wrong here.
- Never use `.then()` or `.catch()` (promise chaining) this is against our coding style conventions. Instead use `async` functions and `await` the promises. If you're using `.then` or `.catch` because it won't work otherwise, it means you're probably doing something you shouldn't. If you need help getting around these situations, let me know and I can help guide you through them

PR #71, opened 2021-07-15
Work/leo/z auction nft approve

SalvoAgustin:
Created PR to feature/colbr/zAuction branch with useApprovals hook


src/lib/hooks/useApprovals.tsx
```diff
@@ -0,0 +1,65 @@
...
+	sendApprovalForAll: (
+		params: ForAllApprovalParams,
+	) => Promise<ethers.ContractTransaction>;
+}
+
+export function useApprovals(): ApproveNFTHook {
+    
+	const registry = useZnsContracts()?.registry;
+
+	const sendSingleApproval = async (params: SingleApprovalParams) => {
```
Remscar, line 28: Please rename to `approveSingleToken`

src/lib/hooks/useApprovals.tsx
```diff
@@ -0,0 +1,65 @@
...
+				params.tokenId,
+			);
+
+			return tx;
+		}, `Send Single Approval`);
+
+		return tx;
+	};
+
+	const sendApprovalForAll = async (params: ForAllApprovalParams) => {
```
Remscar, line 46: Please rename to `approveAllTokens`

src/lib/hooks/useApprovals.tsx
```diff
@@ -0,0 +1,65 @@
...
+				params.approved,
+			);
+
+			return tx;
+		}, `Send Approval for All`);
+
+		return tx;
+	};
+
+	return { sendSingleApproval , sendApprovalForAll };
```
Remscar: Please add 2 additional methods;

`getTokenOperator` (returns what address is approved for a specific token)
`isApprovedForAllTokens` (returns whether or not a specified address has been approved for all tokens)

src/lib/hooks/useApprovals.tsx
```diff
@@ -0,0 +1,61 @@
...
+				params.approved,
+			);
+
+			return tx;
+		}, `Send Approval for All`);
+
+		return tx;
+	};
+
+	return { approveSingleToken, approveAllTokens };
```
Remscar, line 60: Please add 2 additional methods;

getTokenOperator (returns what address is approved for a specific token)
isApprovedForAllTokens (returns whether or not a specified address has been approved for all tokens)

src/lib/hooks/useCheckApprovals.tsx
```diff
@@ -0,0 +1,45 @@
+import { useZnsContracts } from 'lib/contracts';
+import { tryFunction } from 'lib/utils';
+import React from 'react';
+
+interface CheckApproveHook {
+	tokenOperator: string | null;
+	approvedForAllTokens: boolean | null;
+}
+
+export function useApprovals(
```
Remscar, line 10: This was a good way to solve this if we were to use the data from this in a UI component directly, but we most likely will be using these methods in some business logic on-demand. As such, can you please refactor these to be functions, instead of state variables. If we need caching, we can implement that at another layer.

Also please include these methods in the existing `useApprovals.tsx` hook that you made instead of a separate hook like you did here.

PR #74, opened 2021-07-21
Added query function promise returning domain data or undefined

SalvoAgustin:
Created PR for multiple domain data query

### DIFF
--- a/src/components/Tables/RequestTable/RequestTable.tsx
+++ b/src/components/Tables/RequestTable/RequestTable.tsx
@@ -42,6 +42,8 @@ import grid from './assets/grid.svg';
 import list from './assets/list.svg';
 import { useZnsContracts } from 'lib/contracts';
 import { useWeb3React } from '@web3-react/core';
+import { getDomainData } from 'lib/useDomainStore';
+import { useSubgraphProvider } from 'lib/providers/SubgraphProvider';
 
 type RequestTableProps = {
 	style?: React.CSSProperties;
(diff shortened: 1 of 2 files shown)


src/lib/useDomainStore.ts
```diff
@@ -331,3 +339,18 @@ export { useDomainStore };
 // function subdomains(subdomains: any, arg1: string) {
 //   throw new Error('Function not implemented.');
 // }
+
+export const getDomainData = async (
+	domainId: string,
+	client: ApolloClient<NormalizedCacheObject> | undefined,
```
Remscar, line 345: Require that the client exists (is defined) ie change this to just be `client: ApolloClient<NormalizedCacheObject>`

It should be up to the user to make sure that the client is defined, if it isn't, this function can't do anything.

src/components/Tables/RequestTable/RequestTable.tsx
```diff
@@ -226,6 +229,21 @@ const RequestTable: React.FC<RequestTableProps> = ({ style, userId }) => {
 		});
 	}, [yourRequests.requests, requestsForYou.requests, domainFilter]);
 
+	//sample effect to check that the function fetchs the domain data of an example id
+	useEffect(() => {
```
Remscar, line 233: Please remove this sample before we merge this

src/components/Tables/RequestTable/RequestTable.tsx
```diff
@@ -42,6 +42,8 @@ import grid from './assets/grid.svg';
 import list from './assets/list.svg';
```
Remscar, line 42: Please revert all changes to this file since it was for the sample/test code

PR #75, opened 2021-07-23
Work/leo/accept bid refactor

SalvoAgustin:
Created PR, with BidProvider logic to send the acceptBid transaction to zAuctionSupportingZNS contract, and moved useZAuctionHook logic inside


src/lib/types.ts
```diff
@@ -106,5 +108,12 @@ export type Bid = {
 	amount: number;
 	bidderAccount: string;
 	date: Date;
-	tokenId?: string;
+	tokenId: number;
+
+	signature: ethers.utils.BytesLike;
```
Remscar, line 113: You can change this to just `string` to reduce code smell

src/lib/types.ts
```diff
@@ -106,5 +108,12 @@ export type Bid = {
 	amount: number;
 	bidderAccount: string;
 	date: Date;
-	tokenId?: string;
+	tokenId: number;
```
Remscar, line 111: Token ID is a `uint256` which is larger than a JS `number` can hold, this should be `string` instead

src/lib/types.ts
```diff
@@ -106,5 +108,12 @@ export type Bid = {
...
+
+	signature: ethers.utils.BytesLike;
+	auctionId: number;
```
Remscar, line 114: Auction id is also a `uint256` in solidity so it needs to be a string here

src/lib/types.ts
```diff
@@ -106,5 +108,12 @@ export type Bid = {
...
+	nftAddress: string;
+	minBid: number;
```
Remscar, line 116: minBid, startBlock, and expireBlock are all potentially uint256's, so keep these as strings please

src/lib/providers/BidProvider.tsx
```diff
@@ -78,12 +86,27 @@ const BidProvider: React.FC<BidProviderType> = ({ children }) => {
...
+		
+		if (!zAuctionContract) {
+			throw Error(`no contract`);
+		}
 		try {
-			await new Promise((resolve) => setTimeout(resolve, 2000));
-			addNotification(`Bid accepted`);
-			return;
+
+			const tx = await zAuctionContract?.acceptBid(
```
Remscar, line 98: You could use the `tryFunction` helper here that you've used in other hooks to simplify this code

src/lib/providers/BidProvider.tsx
```diff
@@ -113,7 +137,13 @@ const BidProvider: React.FC<BidProviderType> = ({ children }) => {
 						bidderAccount: id,
 						amount,
 						date: new Date(), // not supported by zAuction
-						tokenId: e.tokenId,
+						tokenId: Number(e.tokenId),
```
Remscar, line 140: Bonus points: Consider making a helper function so you don't repeat the same code here that you do in the `getBidsForDomain` this would follow the D.R.Y principle 

The helper could take either DTO's from the endpoints, and return a `Bid`

src/lib/providers/BidProvider.tsx
```diff
@@ -127,6 +146,40 @@ const BidProvider: React.FC<BidProviderType> = ({ children }) => {
 		}
 	};
 
+	//this will receive either DTOs, and will populate the parameters with the correct data
+	function getBidParameters(
+		DTO: NftIdBidsDto | AccountBidsDto,
+		missingString: string,
```
Remscar, line 152: Consider instead having two optional arguments to this function `account` and `tokenId`. You can default to what's in the dto, but if it's undefined, you use the optional argument.

Remscar (approved):

PR #76, opened 2021-07-27
Work/leo/use z auction api

SalvoAgustin:
Created PR for the API refactor of the endpoints and function calls, created a hook to get the apiEndpoint name based on user selected chain, and refactorized zAuction.ts functions to receive the apiEndpoint as argument to fetch the data

### DIFF
--- a/src/lib/hooks/useZAuctionAPI.ts
+++ b/src/lib/hooks/useZAuctionAPI.ts
@@ -17,8 +17,7 @@ export function useZAuctionAPI(chain: number): string {
 			}
 		}
 
-		const apiEndpointName: string = getApiEndpoint();
-		setApi(apiEndpointName);
+		setApi(getApiEndpoint());
 	}, [chain]);
 
 	return api;
--- a/src/lib/hooks/useZAuctionAPI.ts
+++ b/src/lib/hooks/useZAuctionAPI.ts
@@ -1,9 +1,10 @@
 import React, { useState } from 'react';
 
 //the hook that returns the useZauctionAPI with the api endpoint that match with the selected chain
-const [api, setApi] = useState('Unsupported Chain');
 
 export function useZAuctionAPI(chain: number): string {
+	const [api, setApi] = useState('Unsupported Chain');
+
 	React.useEffect(() => {
 		function getApiEndpoint() {
 			switch (chain) {
--- a/src/lib/hooks/useZAuctionAPI.ts
+++ b/src/lib/hooks/useZAuctionAPI.ts
@@ -1,0 +1,25 @@
+import { useChainSelector } from 'lib/providers/ChainSelectorProvider';
+import React, { useState } from 'react';
+
+//the hook that returns the useZauctionAPI with the api endpoint that match with the selected chain
+const [api, setApi] = useState('Unsupported Chain');
+
+export function useZAuctionAPI(chain: number): string {
+	React.useEffect(() => {
+		function getApiEndpoint() {
+			switch (chain) {
+				case 1:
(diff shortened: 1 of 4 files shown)


src/lib/hooks/useZAuctionAPI.ts
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from 'react';
+
+//the hook that returns the useZauctionAPI with the api endpoint that match with the selected chain
+
+export function useZAuctionAPI(chain: number): string {
+	const [api, setApi] = useState('Unsupported Chain');
```
Remscar, line 6: Instead of "Unsupported Chain" this should be `undefined`.

Therefore it would be `useState<string | undefined>` where a string means a valid endpoint, but undefined is an invalid endpoint (unsupported chain)

You will then have to check `api` to ensure that is a `string` (and not undefined) before using it.

src/lib/zAuction.ts
```diff
@@ -57,29 +48,18 @@ interface BidPostInterface {
...
-	let chain = 42; //useChainSelector().selectedChain;
-	switch (chain) {
-		case 1:
-			return 'https://zproduction.ilios.dev/api';
-		case 42:
-			return 'https://zproxy.ilios.dev/api';
-		default:
-			return 'Unsupported Chain';
-	}
+function setApiEndpoints(apiName: string) {
```
Remscar, line 51: Nit: rename to `getApiEndpoints`

A function with `set` in it's name implies you are setting the state of some long-lived object. This function does not write to any state, it returns a newly created object. "Get" is a more proper verb for this sort of action.

Remscar, line 51: `apiName` is a misleading parameter name. You are passing in an URI here, not the name of an API.

Therefore a name like "baseUri` or `baseApiUri` or `BaseApiEndpoint` would be more fitting.

src/lib/hooks/useZAuctionAPI.ts
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from 'react';
+
+//the hook that returns the useZauctionAPI with the api endpoint that match with the selected chain
+
+export function useZAuctionAPI(chain: number): string {
```
Remscar, line 5: Let's rename this to `useZAuctionBaseApiUri`

src/lib/zAuction.ts
```diff
@@ -89,10 +69,14 @@ function getNftId(contract: string, tokenId: string) {
 	return nftId;
 }
 
-export async function getBidsForNft(contract: string, tokenId: string) {
+export async function getBidsForNft(
+	contract: string,
+	tokenId: string,
+	apiName: string,
```
Remscar, line 75: Please rename this to `baseApiUri` instead of `apiName`.

Also; Please move the `baseApiUri` parameter to be the first parameter in all of these function calls.

src/lib/providers/BidProvider.tsx
```diff
@@ -126,13 +130,14 @@ const BidProvider: React.FC<BidProviderType> = ({ children }) => {
 	};
 
 	const getBidsForAccount = async (id: string) => {
+		if (baseApiUri === undefined) {
+			throw Error(`no api endpoint`);
+		}
 		try {
-			const bids = await zAuction.getBidsForAccount(id);
+			const bids = await zAuction.getBidsForAccount(baseApiUri!, id);
```
Remscar, line 137: You shouldn't have to use `!` here if you did the check above to make sure it is not undefined.

Remscar (approved):
1 minor comment, otherwise looks great!

PR #77, opened 2021-07-28
Changed to await until the query finish the data fetch

SalvoAgustin:
Creating PR for the hyperlink fix, now it awaits until the query finish loading


src/pages/ZNS/ZNS.tsx
```diff
@@ -103,6 +104,7 @@ const ZNS: React.FC<ZNSProps> = ({ domain, version }) => {
 
 	// Force to go back to home if invalid domain
 	React.useEffect(() => {
+		if(loading === false){
```
Remscar, line 107: Style nit; This appears to not have been formatted, please make sure to run this through the code styler.

Remscar, line 107: Prefer to use `if (!loading)` instead of `if (loading === false)` since it's a slightly safer operation (there is a subtle different between the two)

Remscar (approved):

PR #81, opened 2021-07-29
Work/leo/api slow load fix

SalvoAgustin:
Created a PR for the API slow down page load, changed table to fetch, display and update container height on each one instead of awaiting for all

### DIFF
--- a/src/components/Tables/DomainTable/DomainTable.tsx
+++ b/src/components/Tables/DomainTable/DomainTable.tsx
@@ -91,8 +91,8 @@ const DomainTable: React.FC<DomainTableProps> = ({
 	const [biddingOn, setBiddingOn] = useState<DisplayDomain | undefined>();
 
 	const [data, setData] = useState<DomainData[]>([]);
-	const [lastDomains, setLastDomains] = useState<Domain[]>();
-	const [domainsChanged, setDomainsChanged] = useState(true);
+	const [tableData, setTableData] = useState<DomainData[]>([]);
+	const [isSubscribed, setIsSubscribed] = useState(false);
 	const containerRef = useRef<HTMLDivElement>(null);
 
 	///////////////
@@ -156,54 +156,7 @@ const DomainTable: React.FC<DomainTableProps> = ({
 
 	useEffect(() => {
 		setIsLoading(true);
-	}, [domainsChanged]);
-
-	// Gets metadata for each NFT in domain list
-	useEffect(() => {
-		const loaded: DomainData[] = [];
-
-		// Get metadata
-		const getData = async (domain: Domain) => {
-			try {
-				const [metadata, bids] = await Promise.all([
-					getMetadata(domain.metadata),
-					getBidsForDomain(domain),
-				]);
-
-				if (!metadata) {
-					console.log(`found no metadata for ${domain.id}`);
-					return;
-				}
-
-				// Filter out user's bids if configured to do so
-				let filteredBids;
-				if (hideOwnBids) {
-					filteredBids = bids?.filter(
-						(bid: Bid) => bid.bidderAccount !== userId,
-					);
-				}
-
-				loaded.push({
-					domain: domain,
-					metadata: metadata,
-					bids: filteredBids || bids || [],
-				});
-
-				setData(loaded);
-				setIsLoading(true); //we trigger here the resize on every new fetch
-				setIsLoading(false);
-				
-				if (onLoad) {
-					onLoad();
-				}
-			} catch (e) {}
-		};
-
-		for (let i = 0; i < domains.length; i++) {
-			if (!domains[i].metadata) continue;
-			getData(domains[i]);
-		}
-	}, [domainsChanged]);
+	}, [domains]);
 
 	// Resizes the table container
 	// (The animation is done in CSS)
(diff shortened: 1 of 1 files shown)


src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -175,73 +173,67 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
-			getData(domains[i]);
-		}
-
-		if (!count) {
-			setHasMetadataLoaded(true);
-			setIsLoading(false);
-		}
-	}, [domains]);
+		getData(domains);
+	}, [domainsChanged]); //this was [domains], but that way it triggers a weird bug where the page re render and fetch bad
```
Remscar, line 218: This bug is most likely caused due to an outdated hook. You need to make sure that this effect is cleaned up after a previous render if it hasn't completed yet: https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

This is accomplished using the `isSubscribed` pattern you can find in some other hooks in this project: https://github.com/zer0-os/zNS-dapp/blob/c685f0d795ec31c3d3eb66506aeff53d210f2250/src/components/Profile/Profile.tsx#L86

Essentially, you should never call a `set____` (React set state) function unless you are sure the hook is still valid. This is especially important for long running operations such as this. If for some reason you are no longer subscribed, you just exit out.

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -175,73 +173,67 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
 	useEffect(() => {
-		const loaded: DomainData[] = [];
-		setHasMetadataLoaded(false);
-		let count = 0;
-		let completed = 0;
+		let loaded: DomainData[] = [];
 
 		// Get metadata
-		const getData = async (domain: Domain) => {
+		const getData = async (domain: Domain[]) => {
```
Remscar, line 179: So it looks like what you've actually done is switch this from fetching all of the domain data in parallel, to instead making it fetch domain data synchronously.

This will actually slow down the total page load time, but make it slightly quicker to show at least one ui element.

Try to take another crack at this. I'm not sure if you're familiar with race conditions, but it's very likely that a race condition could pop up here (when setting `loadedDomains`), in which case you would need to employ a mutex.

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -190,11 +180,20 @@ const DomainTable: React.FC<DomainTableProps> = ({
 
 				completed++;
 
+				if (completed === domains.length) {
+					setIsSubscribed(false);
```
Remscar, line 184: In this case, the `isSubscribed` variable should *not* be a React state variable. It is a local variable only to the Effect, and should be defined above. (i will mark it)

Make sure you look at the example code I linked earlier; https://github.com/zer0-os/zNS-dapp/blob/c685f0d795ec31c3d3eb66506aeff53d210f2250/src/components/Profile/Profile.tsx#L86

In the example, isSubscribed is not a state variable, it shouldn't be one here either.

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -203,56 +202,69 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
 	useEffect(() => {
 		if (!isLoading && onLoad) onLoad();
 	}, [isLoading]);
 
+	/////////////////////
+	///// Functions /////
+	/////////////////////
+
+	const feed = () => {
+		let cachedData = tableData; //last data feeded
```
Remscar, line 238: Make sure to use `const` and not `let` if you aren't re-defining variables.

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -203,56 +202,69 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
+
+		let lengthDiff = data.length - cachedData.length
+		let feetIteration =  lengthDiff < chunkLength ? lengthDiff : chunkLength //how many data we need to feed, based on the fetched data we have
```
Remscar, line 241: `feetIteration` -> `feedIteration`?

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -203,56 +202,69 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
+		
+		for (
+			let i = cachedData.length;
+			i < feetIteration;
+			i++ //adds new chunk of 5 domains
+		)
+			cachedData.push(data[i]);
```
Remscar, line 248: Surround `for` loop code with brackets, don't use indentation shortcuts like this

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -203,56 +202,69 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
+	const feed = () => {
+		let cachedData = tableData; //last data feeded
+
+		let lengthDiff = data.length - cachedData.length
+		let feetIteration =  lengthDiff < chunkLength ? lengthDiff : chunkLength //how many data we need to feed, based on the fetched data we have
+		
+		for (
+			let i = cachedData.length;
+			i < feetIteration;
+			i++ //adds new chunk of 5 domains
```
Remscar, line 246: Remove comments like this, the 5 refers to the setting of `chunkLength` but could be changed and then this comment is outdated and confusing.

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -81,21 +81,17 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
 	const [isLoading, setIsLoading] = useState(true);
-
 	const [containerHeight, setContainerHeight] = useState(0);
 	const [searchQuery, setSearchQuery] = useState('');
-
 	const [modal, setModal] = useState<Modals | undefined>();
 	const [biddingOn, setBiddingOn] = useState<DisplayDomain | undefined>();
-
-	const [loadedDomains, setLoadedDomains] = useState<DomainData[]>([]);
+	const [data, setData] = useState<DomainData[]>([]);
```
Remscar, line 89: `data` is incredibly vague. All variables are data. But what is this variable actually? What does it represent?

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -81,21 +81,17 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
+	const [tableData, setTableData] = useState<DomainData[]>([]);
+	const [isSubscribed, setIsSubscribed] = useState(false);
 
 	const containerRef = useRef<HTMLDivElement>(null);
-
+	const chunkLength = 5;
```
Remscar, line 94: Why does chunk length matter, leave a comment here to describe the purpose of this variable

src/components/Tables/DomainTable/DomainTable.tsx
```diff
@@ -203,56 +202,69 @@ const DomainTable: React.FC<DomainTableProps> = ({
...
 
 	useEffect(() => {
 		if (!isLoading && onLoad) onLoad();
 	}, [isLoading]);
 
+	/////////////////////
+	///// Functions /////
+	/////////////////////
+
+	const feed = () => {
```
Remscar, line 237: Where/when does feed get called?

src/components/Tables/DomainTable/DomainTable.tsx
Remscar: Also why is it only being added in chunks?

src/components/Tables/DomainTable/DomainTable.tsx
SalvoAgustin: That function is for Brett, he needs to test it with UI changes before continue

src/components/Tables/DomainTable/DomainTable.tsx
SalvoAgustin: Changed to only load one first chunk, right now first 5 domains. Brett will call feed() later and in each call it adds one domain more to the table, so he can load on demand instead of chunks

Remscar:
Closing this PR since Brett is going to try to tackle this a bit differently.

PR #84, opened 2021-08-06
Work/leo/bid list endpoint

SalvoAgustin:
Created PR with a new function exposed at zAuction, to fetch an array of bids list, of the NFTIds array sended

### DIFF
--- a/src/lib/zAuction.ts
+++ b/src/lib/zAuction.ts
@@ -8,7 +8,16 @@ export interface NftIdBidsDto {
 	minimumBid: string;
 	startBlock: string;
 	expireBlock: string;
-	date: string;
+}
+
+interface NftIdBidsAtListDto {
+	account: string;
+	auctionId: string;
+	bidAmount: string;
+	expireBlock: string;
+	minimumBid: string;
+	signedMessage: string;
+	startBlock: string;
 }
 
 export interface AccountBidsDto {
@@ -38,8 +47,10 @@ interface CreateBidDto {
 	nftId: string;
 }
 
-interface BidsListDto {
-	bids: string[]
+interface BidsListResponseDto {
+	bids: NftIdBidsAtListDto[];
+	contractAddress: string;
+	tokenId: string;
 }
 
 interface BidPostInterface {
@@ -97,7 +108,6 @@ export async function getBidsListForNfts(
 	tokenId: string[],
 ) {
 	const nftIdsArray: string[] = [];
-	const bidsList: BidsListDto[] = [];
 
 	for (let i = 0; i < tokenId.length; i++) {
 		// it builds the array of nftIds strings
@@ -114,18 +124,21 @@ export async function getBidsListForNfts(
 		body: JSON.stringify(nftIdsBody),
 	});
 
-	const jsonResponse = await response.json(); //awaits the json of the response
+	const jsonResponse = (await response.json()) as BidsListResponseDto[]; //awaits the json of the response
+
+	const bidsList = new Map();
 
 	for (let i = 0; i < tokenId.length; i++) {
-		//builds an array of only bid lists
-		bidsList.push(jsonResponse[i].bids);
+		// it builds the array of nftIds strings
+		bidsList.set(jsonResponse[i].tokenId, jsonResponse[i].bids);
 	}
 
+	console.log(bidsList);
 	return bidsList;
 }
 
 export async function getBidsForAccount(baseApiUri: string, id: string) {
-	let endpoints = getApiEndpoints(baseApiUri);
+	const endpoints = getApiEndpoints(baseApiUri);
 	const response = await fetch(endpoints.accountBidsEndpoint + id);
 	const bids = (await response.json()) as AccountBidsDto[];
 	return bids;
@@ -138,7 +151,7 @@ async function encodeBid(
 	if (!ethers.utils.isAddress(bid.contractAddress)) {
 		throw Error(`Invalid contract address ${bid.contractAddress}`);
 	}
-	let endpoints = getApiEndpoints(baseApiUri);
+	const endpoints = getApiEndpoints(baseApiUri);
 	const response = await fetch(endpoints.encodeBidEndpoint, {
 		method: 'POST',
 		headers: { 'Content-Type': 'application/json' },
@@ -157,7 +170,7 @@ async function sendBid(
 	if (!ethers.utils.isAddress(bid.contractAddress)) {
 		throw Error(`Invalid contract address ${bid.contractAddress}`);
 	}
-	let endpoints = getApiEndpoints(baseApiUri);
+	const endpoints = getApiEndpoints(baseApiUri);
 	await fetch(`${endpoints.bidsEndpoint}${nftId}`, {
 		method: 'POST',
 		headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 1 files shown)


src/lib/zAuction.ts
```diff
@@ -87,6 +87,39 @@ export async function getBidsForNft(
...
+
+	let endpoints = getApiEndpoints(baseApiUri);
+
+	const response = await fetch(endpoints.bidListEndpoint, {
+		method: 'POST',
+		headers: { 'Content-Type': 'application/json' },
+		body: JSON.stringify(nftIdsBody),
+	});
+
+	const jsonResponse = await response.json(); //awaits the json of the response
```
Remscar, line 113: You should create an interface and type this to that interface. Currently `jsonResponse` is `any` which we want to avoid.

src/lib/zAuction.ts
```diff
@@ -87,6 +87,39 @@ export async function getBidsForNft(
...
+
+	for (let i = 0; i < tokenId.length; i++) {
+		//builds an array of only bid lists
+		bidsList.push(jsonResponse[i].bids);
+	}
+
+	return bidsList;
```
Remscar, line 120: The return of this function is `any[]` which we want to avoid.

src/lib/zAuction.ts
```diff
@@ -87,6 +87,39 @@ export async function getBidsForNft(
...
+		method: 'POST',
+		headers: { 'Content-Type': 'application/json' },
+		body: JSON.stringify(nftIdsBody),
+	});
+
+	const jsonResponse = await response.json(); //awaits the json of the response
+
+	for (let i = 0; i < tokenId.length; i++) {
+		//builds an array of only bid lists
+		bidsList.push(jsonResponse[i].bids);
```
Remscar, line 117: By the logic of your function `bidsList` is a `string[]` and you are pushing `jsonResponse[i].bids` into the array of string...

Therefore you'd think that `jsonResponse[i].bids` is a string, but I'm almost certain it is not.

src/lib/zAuction.ts
```diff
@@ -87,6 +87,39 @@ export async function getBidsForNft(
 	return bids;
 }
 
+export async function getBidsListForNfts(
```
Remscar, line 90: Think about what is going to be easiest for a user to use.

In our application is there any case where a user is going to want to get bids from multiple NFT's from multiple contracts? We only are ever dealing with one contract per network, but your interface asks for an array of contracts.

This makes it cumbersome for a user to use since if they want to get bids from 100 different tokens, they're going to have to fill a second array with 100 of the same value.

I recommend you make the contract input a `contract: string` and just re-use the contract.

src/lib/zAuction.ts
```diff
@@ -87,6 +87,39 @@ export async function getBidsForNft(
...
+	const bidsList: string[] = [];
+
+	for (let i = 0; i < tokenId.length; i++) {
+		// it builds the array of nftIds strings
+		nftIdsArray.push(getNftId(contract[i], tokenId[i]));
+	}
+
+	const nftIdsBody = { nftIds: nftIdsArray }; //builds the correct body to send
+
+	let endpoints = getApiEndpoints(baseApiUri);
```
Remscar, line 105: This should be const not let.

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
...
+
+	const response = await fetch(endpoints.bidListEndpoint, {
+		method: 'POST',
+		headers: { 'Content-Type': 'application/json' },
+		body: JSON.stringify(nftIdsBody),
+	});
+
+	const jsonResponse = (await response.json()) as BidsListResponseDto[]; //awaits the json of the response
+
+	const bidsListMap = new Map <string, NftIdBidsOfListsDto[]>(); //mapping of tokenId as key and returned bids as value
```
Remscar, line 130: The Typescript implementation of `Map<>` can be kinda heavy handed here. You can use a simpler map instead by creating an interface as such:

```
interface TokenBids {
  [tokenId: string]: Bid[] | undefined
 }
 
 const bidsMap = {} as TokenBids;
 bidsMap[someTokenId] = { some bid }
 ```

File: src/lib/zAuction.ts [hunk 7]
```diff
@@ -11,6 +11,16 @@ export interface NftIdBidsDto {
 	date: string;
 }
 
+export interface NftIdBidsOfListsDto {
```
Remscar, line 14: This is a bit of a confusing interface name; perhaps you could make it more simple by calling it `NftBidDto`

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
...
+		method: 'POST',
+		headers: { 'Content-Type': 'application/json' },
+		body: JSON.stringify(nftIdsBody),
+	});
+
+	const jsonResponse = (await response.json()) as BidsListResponseDto[]; //awaits the json of the response
+
+	const bidsListMap = new Map <string, NftIdBidsOfListsDto[]>(); //mapping of tokenId as key and returned bids as value
+
+	for (let i = 0; i < tokenId.length; i++) { //populate the map with the data for every user
```
Remscar, line 132: You are iterating the `jsonResponse` array but using the `tokenId` array as the iterator. This is dangerous. There may be a bug in the zAuction API where `jsonResponse.length !== tokenId.length` and this code will break

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
 	return bids;
 }
 
+export async function getBidsListForNfts(
+	baseApiUri: string,
+	contract: string,
+	tokenId: string[],
```
Remscar, line 109: `tokenId` implies a singular token id, but this is an array, thus `tokenIds` would be more appropriate here.

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
...
+) {
+	const nftIdsArray: string[] = [];
```
Remscar, line 111: Putting `Array` in variable names is redundant in Typescript since you can imply that something is an array based on it it's type. Having a name like `nftIds` would be wiser here since the user can easily see that it is a `string[]`

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
...
+
+	for (let i = 0; i < tokenId.length; i++) {
```
Remscar, line 113: You are also iterating `nftIdsArray` but with `tokenId` instead. Never iterate Array A with start/stop from array B.

src/lib/zAuction.ts
```diff
@@ -87,8 +103,41 @@ export async function getBidsForNft(
...
+		// it builds the array of nftIds strings
+		nftIdsArray.push(getNftId(contract, tokenId[i]));
+	}
+
+	const nftIdsBody = { nftIds: nftIdsArray }; //builds the correct body to send
```
Remscar, line 118: Nitpick: `nftIdsBody` -> `requestBody` since what you are doing here is building the body for the HTTP **request** you are about to make.

Remscar (approved):
Well done. I love the changes to use `.map` and `.forEach`

PR #85, opened 2021-08-09
Work/leo/accept preview image

SalvoAgustin:
Created PR for the preview image optional in metadata


src/lib/utils/index.ts
```diff
@@ -19,22 +20,45 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
-
+const getMetadataObject = async (params: DomainMetadataParams) => {
+	// upload images to http backend
 	const imageResponse = await fetch(uploadApiEndpoint, {
 		method: 'POST',
 		body: JSON.stringify(params.image),
 	});
 	const image = (await imageResponse.json()) as UploadResponseDTO;
 
+	if (params.previewImage!) {
```
Remscar, line 31: `params.previewImage` may be null here. Please do not use `!` like this (non-null assertion operator) unless absolutely necessary, it's a generally dangerous operator to use.

Instead a better way to do this check is:

```
if (params.previewImage) {
...
}
```

src/lib/utils/index.ts
```diff
@@ -19,22 +20,45 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
+
+		const metadataObject = {
+			name: params.name,
+			description: params.story,
+			image: image.url,
+			previewImage: previewImage.url,
+		};
+
+		return metadataObject;
+	} else {
```
Remscar, line 48: Refactor this if/else statement so that there is no `else` statement.

src/lib/utils/index.ts
```diff
@@ -19,22 +20,45 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
+};
+
+export const createDomainMetadata = async (params: DomainMetadataParams) => {
 	// upload metadata to IPFS
-	const metadataObject = {
-		name: params.name,
-		description: params.story,
-		image: image.url,
-	};
+	const metadataObject = await getMetadataObject(params);
```
Remscar, line 61: Well done creating a helper method to do this logic.

src/lib/utils/index.ts
```diff
@@ -19,22 +20,45 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
  * @param params Metadata parameters
  * @returns URI to the created Metadata
  */
-
-export const createDomainMetadata = async (params: DomainMetadataParams) => {
-	// upload image to http backend
-
+const getMetadataObject = async (params: DomainMetadataParams) => {
```
Remscar, line 23: Consider renaming this helper method to something along the lines of `uploadMetadata` or something similar.

`getMetadataObject` implies that you are fetching or retrieving a stored value, when this method is actually `transforming` (data), and `uploading` (data).

src/lib/utils/index.ts
```diff
@@ -1,11 +1,19 @@
 export * from './domains';
 
 interface DomainMetadataParams {
+	previewImage?: Buffer;
 	image: Buffer;
 	name: string;
 	story: string;
 }
 
+interface MetadataObject {
```
Remscar, line 10: It was smart to create a helper object here, however an object that fulfills this purpose already exists. If you look in `types.ts` https://github.com/zer0-os/zNS-dapp/blob/3e4f579e814e843981a7220b0a151fc34e04a899/src/lib/types.ts#L20

You will see a `Metadata` interface that gets declared. This is the interface which **all** metadata objects should follow.

Please use that interface instead of creating your own.

Additionally; with this object you have created, you have introduced a bug into the dApp. You have renamed `image` to `imageUri`.
The `Metadata` interface I mentioned is the interface which all metadata objects get casted as when they are **read** from IPFS. The code you are writing is the code that **writes** to IPFS. Read/Write operations to the same data should share the same interface (as mentioned above) but when they don't, bugs like this can happen where you write one set of data, but then try to read a different set of data. So if this code was used, all images would be broken on read.

src/lib/utils/index.ts
```diff
@@ -19,22 +27,38 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
+	if (params.previewImage) {
+		//if params has the optional preview image it builds metadata with it
+		const previewImageResponse = await fetch(uploadApiEndpoint, {
+			method: 'POST',
+			body: JSON.stringify(params.previewImage),
+		});
+		const previewImage =
+			(await previewImageResponse.json()) as UploadResponseDTO;
+		metadataObject.previewImageUri = previewImage.url;
+		return metadataObject;
```
Remscar, line 53: Is this return statement needed? Or can it be removed and the code will function exactly the same?

src/lib/utils/index.ts
```diff
@@ -19,22 +27,38 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
+	let metadataObject: MetadataObject = {
 		name: params.name,
 		description: params.story,
-		image: image.url,
+		imageUri: image.url,
 	};
+
+	if (params.previewImage) {
+		//if params has the optional preview image it builds metadata with it
+		const previewImageResponse = await fetch(uploadApiEndpoint, {
```
Remscar, line 46: This segment of code:

```
const previewImageResponse = await fetch(uploadApiEndpoint, {
	method: 'POST',
	body: JSON.stringify(params.previewImage), // only difference
});
```

is repeated twice in this method. Consider refactoring this into a helper method.

src/lib/utils/index.ts
```diff
@@ -19,22 +27,38 @@ const uploadApiEndpoint = `https://zns-backend.netlify.app/.netlify/functions/up
...
+
+	return metadataObject;
+};
+
+export const createDomainMetadata = async (params: DomainMetadataParams) => {
+	// upload metadata to IPFS
+	const metadataObject = await uploadMetadata(params);
 	const metadataAsString = JSON.stringify(metadataObject);
 
 	const metadataResponse = await fetch(uploadApiEndpoint, {
```
Remscar, line 64: Consider also refactoring this to use that helper method i mentioned earlier.

Remscar (commented):
Well done. This is good code. Please resolve the merge conflicts and then it will be good to merge.

Remscar (approved):

PR #102, opened 2021-08-15
LOOT Token Integration

SalvoAgustin:
$LOOT integration to domain request flow


Remscar (approved):
We will replace mainnet addresses later this week

PR #103, opened 2021-08-16
Fixes for problems that arise when switching/disconnecting wallets

colbr:
**Problem 1**
Problem: Used to be able to bid on your own domains by opening the bid modal with a different wallet connected, then switch to your main wallet.
Solution: Flow already resets on wallet change - added a check on the first step of the flow to disable the input/continue button when `connected wallet ID === domain owner ID`.

**Problem 2**
Problem: Should not be able to have bid or profile modals open without a wallet connected. With modals open, disconnecting wallet would not close them, which could lead to some functionality crashing.
Solution: Added checks for connected account - if not connected close the bid or profile modal (if open). Also made the bid and profile modals render blank if no `userId` is supplied - not sure if this is the greatest solution.

**Problem 3**
Problem: 'Make A Bid' button should be disabled when no wallet is connected.
Solution: Disabled 'Make A Bid' when no wallet is connected

### DIFF
--- a/src/pages/ZNS/ZNS.tsx
+++ b/src/pages/ZNS/ZNS.tsx
@@ -257,6 +257,15 @@ const ZNS: React.FC<ZNSProps> = ({ domain, version, isNftView: nftView }) => {
 		}
 		if (triedEagerConnect)
 			addNotification(active ? 'Wallet connected.' : 'Wallet disconnected.');
+
+		// Check if we need to close a modal
+		if (
+			(!active && modal === Modal.Profile) ||
+			modal === Modal.Transfer ||
+			modal === Modal.Mint
+		) {
+			closeModal();
+		}
 	}, [active]);
 
 	useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/pages/ZNS/ZNS.tsx
```diff
@@ -257,6 +257,15 @@ const ZNS: React.FC<ZNSProps> = ({ domain, version, isNftView: nftView }) => {
 		}
 		if (triedEagerConnect)
 			addNotification(active ? 'Wallet connected.' : 'Wallet disconnected.');
+
+		// Check if we need to close a modal
+		if (
+			(!active && modal === Modal.Profile) ||
+			modal === Modal.Transfer ||
```
Remscar, line 264: This logic will close the modal if the transfer or mint modal is open, regardless of `active`. Is this the intended behavior?

(!active && (modal === Modal.Profile || modal === Modal.Transfer || modal === Modal.Mint)) makes more sense here.

PR #106, opened 2021-08-17
Added loading indicators when fullfiling / approving

SalvoAgustin:
Button display after push, a spinner and "fulfilling" or "approving"

### DIFF
--- a/src/components/Confirmation/Confirmation.tsx
+++ b/src/components/Confirmation/Confirmation.tsx
@@ -1,6 +1,7 @@
 //- React Imports
 import { FutureButton } from 'components';
 import LoadingIndicator from 'components/LoadingIndicator/LoadingIndicator';
+import { Maybe } from 'lib/types';
 import React from 'react';
 
 //- Style Imports
@@ -9,6 +10,7 @@ import styles from './Confirmation.module.css';
 type ConfirmationProps = {
 	cancelText?: string;
 	children?: React.ReactNode;
+	errorText?: string;
 	confirmText?: string;
 	showLoading?: boolean;
 	loadingText?: string;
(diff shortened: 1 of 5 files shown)


Remscar:
I should've been a bit more descriptive when making this task. Could you please make it so the loading indicator for approving the staking controller to spend LOOT tokens is on this modal:

https://i.imgur.com/bZS7Vrw.png

And instead of making the `Continue` button have the loading indicator, please make it have the buttons disappear and only the loading indicator is there.

Additionally, you had the loading indicator disappear after the user accepts the transaction in the wallet. The loading state needs to continue until the transaction has been completed. You can get a Promise that only completes when a transaction is completed by called `.wait()` on a transaction receipt.

Follow a flow similar to how approving zAuction to spend WILD tokens to place a bid does it:

- User asked to approve zAuction
- User presses button to approve
- **loading indicator**
- > After user accepts transaction to approve
- **loading indicator continues**
- > After the approve transaction completes
- Flow continues (in this case the user is able to do a transaction to mint an NFT)

Remscar:
Getting a lot closer; https://drive.google.com/file/d/1fUZO74CPU3EEDibMOu7dFOVr3yrucKQ0/view?usp=drivesdk

No need to block the user on the 2nd transaction when the NFT is actually being minted.

The status text needs to change though, instead of just saying "APPROVING" it should say "Please confirm transaction in wallet" and then "Waiting for transaction to complete"

As for the Fulfill status text, just "Please confirm transaction in wallet" works, then when they confirm, we can pop them out

src/components/Confirmation/Confirmation.tsx
```diff
@@ -50,6 +70,16 @@ const Confirmation: React.FC<ConfirmationProps> = ({
 					</FutureButton>
 				</div>
 			)}
+			{!hideButtons && !showLoading && errorText &&(
+			<div>{errorMessage}</div>)}
+			{!hideButtons && showLoading && loadingText && (
+				<div className={styles.Buttons}>
+					<LoadingIndicator
+						style={{ textTransform: 'uppercase' }}
```
Remscar, line 78: Normal case would ideal for this instead of uppercase

Remscar (approved):

PR #123, opened 2021-08-19
Wrapping inactive zNA span in Mint Preview 

colbr:
Per title


src/components/MintPreview/MintPreview.module.css
```diff
@@ -71,6 +71,10 @@
 	font-weight: 700;
 }
 
+.Info span {
```
Remscar, line 74: Will break-word break on periods?

Remscar (approved):

PR #131, opened 2021-08-21
Metamask connecting/reconnecting stuck forever fix

SalvoAgustin:
This error its generated by the lost of authorization to connect, page shouldnt try to reconnect if user isnt authorized yet

Also Metamask may get stuck with requestAccounts promise if user close login overlay, for this, page will try to detect if its stuck and reload, or ask user to reload with a timeout if promise cant resolve. Other DAPPS like opensea / pancakewap / hypertext, displays an error message and ask user to reload

Flow:
User has metamask extension? works OK
User has brave and load crypto wallets at start enabled? works OK

User has brave and not enabled crypto wallets or metamask extension?
-First pop up message asking for extension or crypto wallets to start and reload
-Second when user has the provider available pop ups log in overlay
-Last, if log in sucess, then reloads if necessary and its OK

### DIFF
--- a/src/components/ConnectToWallet/ConnectToWallet.tsx
+++ b/src/components/ConnectToWallet/ConnectToWallet.tsx
@@ -1,5 +1,5 @@
 //- React Imports
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 
 //- Web3 Imports
 import { Web3Provider } from '@ethersproject/providers/lib/web3-provider';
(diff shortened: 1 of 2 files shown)


SalvoAgustin:
Found a bug where if user close connection overlay will get stuck and needs to reload manually, I ll try to detect that and reload automatically

SalvoAgustin:
If user tries to connect metamask, if activate promise solves with requestAccounts stuck error, reloads.
If never resolves due to bug or log in overlay closed, reloads

SalvoAgustin:
In the worst case, if the connect promise wont resolve, and user its stuck, it displays a message asking the user to reload after 20 seconds, let me know if this time its good and messages too

Remscar (approved):

PR #132, opened 2021-08-23
If domain URL its invalid get user back to home after 2 seconds

SalvoAgustin:
If the domain keeps null ( without fetched data ) after some time then its invalid and must redirect user to home


SalvoAgustin:
Let me know if this way sounds good, and if the duration of the timeout sounds good

Remscar (approved):

PR #133, opened 2021-08-24
Coingecko should not render USD $WILD values if fails

SalvoAgustin:
Page wont render $WILD USD values if fails, and values are $0 in that case

### DIFF
--- a/src/lib/providers/CurrencyProvider.tsx
+++ b/src/lib/providers/CurrencyProvider.tsx
@@ -33,7 +33,6 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
 			setWildPriceUsd(price);
 			await new Promise((resolve) => setTimeout(resolve, 30000));
 		} catch (e) {
-			setWildPriceUsd(0);
 			console.error(e);
 			await new Promise((resolve) => setTimeout(resolve, 5000));
 		}
@@ -46,7 +45,6 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
 			setLootPriceUsd(price);
 			await new Promise((resolve) => setTimeout(resolve, 30000));
 		} catch (e) {
-			setLootPriceUsd(0);
 			console.error(e);
 			await new Promise((resolve) => setTimeout(resolve, 5000));
 		}
--- a/src/lib/providers/CurrencyProvider.tsx
+++ b/src/lib/providers/CurrencyProvider.tsx
@@ -29,14 +29,18 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
 	// Poll CoinGecko every 30 seconds
 	const getWildPriceUsd = async () => {
 		const price = await wildTokenPrice();
-		setWildPriceUsd(price || 0);
+		setWildPriceUsd(price);
+		if (price === 0)
+			console.error('$WILD price API call failed, set to $0 by default');
 		await new Promise((resolve) => setTimeout(resolve, 30000));
 		getWildPriceUsd();
 	};
 
 	const getLootPriceUsd = async () => {
 		const price = await lootTokenPrice();
-		setLootPriceUsd(price || 0);
+		setLootPriceUsd(price);
+		if (price === 0)
+			console.error('$LOOT price API call failed, set to $0 by default');
 		await new Promise((resolve) => setTimeout(resolve, 30000));
 		getLootPriceUsd();
 	};
--- a/src/lib/providers/CurrencyProvider.tsx
+++ b/src/lib/providers/CurrencyProvider.tsx
@@ -33,10 +33,10 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
 			setWildPriceUsd(price);
 			await new Promise((resolve) => setTimeout(resolve, 30000));
 		} catch (e) {
-			console.error(e);
 			setWildPriceUsd(0);
+			console.error(e);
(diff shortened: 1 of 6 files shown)


src/lib/tokenPrices.ts
```diff
@@ -5,7 +5,9 @@ import { Maybe } from './types';
 
 const tokenToUsdCache: { [token: string]: number | undefined } = {};
 
-export const tokenToUsd = async (token: string): Promise<number> => {
+export const tokenToUsd = async (
+	token: string,
+): Promise<number | undefined> => {
```
Remscar, line 10: I don't like how this can potentially return `undefined`
If this function fails for some reason, it should throw an error, not return `undefined`.

SalvoAgustin: Understood, lets say that fails to retrieve data, should console the error and not display the value? ( $0 in that case )

src/lib/providers/CurrencyProvider.tsx
```diff
@@ -29,14 +29,16 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
 	// Poll CoinGecko every 30 seconds
 	const getWildPriceUsd = async () => {
 		const price = await wildTokenPrice();
-		setWildPriceUsd(price || 0);
+		setWildPriceUsd(price);
+		if (price === 0) console.error('$WILD price API call failed');
```
Remscar, line 33: Always use brackets when making if statements please.

src/lib/tokenPrices.ts
```diff
@@ -73,6 +73,7 @@ const getLootPrice = async () => {
 	);
 
 	const ethToUsd = await tokenToUsd('ethereum');
+	if (!ethToUsd) return;
```
Remscar, line 76: You've changed the return type of this function from `Promise<number>` to `Promise<number | undefined>` which isn't okay. if `tokenToUsd` fails, it should throw an exception, not 0 or undefined.

src/lib/tokenPrices.ts
```diff
@@ -14,15 +14,15 @@ export const tokenToUsd = async (token: string): Promise<number> => {
...
 		const res = await fetch(
 			`https://api.coingecko.com/api/v3/simple/price?ids=${token}&vs_currencies=usd`,
 		);
 
 		const data = await res.json();
-		if (!data) {
-			throw Error(`Unable to fetch price for ${token}`);
-		}
+
+		if (!data[token]) return 0;
```
Remscar, line 25: This function should throw an exception if it fails, not a value of zero

src/lib/tokenPrices.ts
```diff
@@ -14,15 +14,15 @@ export const tokenToUsd = async (token: string): Promise<number> => {
 
 	if (token === 'LOOT') {
 		priceInUsd = await getLootPrice();
+		if (!priceInUsd) return 0;
```
Remscar, line 17: This function should throw an exception if it fails, not zero

src/lib/providers/CurrencyProvider.tsx
```diff
@@ -29,14 +29,16 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
...
 		await new Promise((resolve) => setTimeout(resolve, 30000));
 		getWildPriceUsd();
 	};
 
 	const getLootPriceUsd = async () => {
 		const price = await lootTokenPrice();
-		setLootPriceUsd(price || 0);
+		setLootPriceUsd(price);
+		if (price === 0) console.error('$LOOT price API call failed');
```
Remscar, line 41: You should be try/catching `await lootTokenPrice()` to see if it fails, not by checking if the price is zero.

src/lib/providers/CurrencyProvider.tsx
Remscar: You should be try/catching await lootTokenPrice() to see if it fails, not by checking if the price is zero.

src/lib/providers/CurrencyProvider.tsx
```diff
@@ -26,18 +26,30 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
...
 	const getWildPriceUsd = async () => {
-		const price = await wildTokenPrice();
-		setWildPriceUsd(price || 0);
-		await new Promise((resolve) => setTimeout(resolve, 30000));
+		try {
+			const price = await wildTokenPrice();
+			setWildPriceUsd(price);
+			await new Promise((resolve) => setTimeout(resolve, 30000));
+		} catch (e) {
+			setWildPriceUsd(0);
```
Remscar, line 36: No need to reset the price to zero. Just leave it as the value it was previously (which will either be undefined or a valid amount)

src/lib/providers/CurrencyProvider.tsx
```diff
@@ -26,18 +26,30 @@ const CurrencyProvider: React.FC<CurrencyProviderType> = ({ children }) => {
...
 	const getLootPriceUsd = async () => {
-		const price = await lootTokenPrice();
-		setLootPriceUsd(price || 0);
-		await new Promise((resolve) => setTimeout(resolve, 30000));
+		try {
+			const price = await lootTokenPrice();
+			setLootPriceUsd(price);
+			await new Promise((resolve) => setTimeout(resolve, 30000));
+		} catch (e) {
+			setLootPriceUsd(0);
```
Remscar, line 49: Same comment as above, don't reset to zero on error.

Remscar (approved):

PR #134, opened 2021-08-25
Bid history should show your own bids

SalvoAgustin:
Changed filter to be optional, and setted from domainTable component props.
Bid history on nftview will display and use the data for all bids, including the ones from the current owner of the domain.
My domains tab in profile, will show only the bids from other users

### DIFF
--- a/src/lib/providers/BidProvider.tsx
+++ b/src/lib/providers/BidProvider.tsx
@@ -17,7 +17,7 @@ import { useChainSelector } from './ChainSelectorProvider';
 export const BidContext = React.createContext({
 	getBidsForDomain: async (
 		domain: Domain,
-		filterOwnerBids?: boolean,
+		filterOwnBids?: boolean,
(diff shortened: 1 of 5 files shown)


src/lib/providers/BidProvider.tsx
```diff
@@ -186,9 +189,10 @@ const BidProvider: React.FC<BidProviderType> = ({ children }) => {
 			)) as zAuction.BidDto[];
 
 			try {
-				bids = bids.filter((e) => {
-					return e.account.toLowerCase() !== domain.owner.id.toLowerCase();
-				});
+				if (filterOwnBids)
```
Remscar, line 192: Minor nitpick;
Please use brackets on if statements, even if it's one statement;

ie:
```
if (filterOwnBids) {
... code
}
```

Remscar (approved):
One single comment that needs to be changed

PR #144, opened 2021-08-28
Fixed the bug that caused $0.00 to show up as the highest bid

Remscar:


colbr (approved):
Looks good @Remscar, clean fix

PR #148, opened 2021-08-31
Changed to display error notifications if subgraph fails with error

SalvoAgustin:
Changed to display errors in zns home , nftView, myDomains and offers at profile. Task card:
https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/8c12bc1c-ca83-4e27-b784-5228e07f4768


colbr:
dApp is about to switch to fully using SDK, so subgraph errors won't be thrown and we likely won't need this feature. Otherwise this was looking good, thanks for putting it together Leo 😊

PR #152, opened 2021-09-02
Back button should only navigate up the domain tree

colbr:
https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/bd639275-0af9-4259-a520-b35cee0841e0

Previously, the back button navigated to the domain you were last one. This isn't the behaviour we want. We want the back button to navigate one level up the tree up the tree.

Solution:
- Check if there's dots in the current domain (i.e. we're at a subdomain)
- Get the "section" of the domain prior to that, (e.g. `/wilder.github.colbr`becomes `/wilder.github` & `/wilder` becomes '/')
- Go to that domain
- Disable back button if we're at the top domain (`/`)

QA Instructions:
https://www.loom.com/share/76a7177fd80048ad91871c0157d7ffb3


stacy-hunt (approved):
Yep sorted!

PR #154, opened 2021-09-02
FIX Blank overlay if user press "make a bid" button disconnected

SalvoAgustin:
Steps to reproduce:
Disconnect wallet
Navigate to NFT View of any domain
Hit "Make A Bid"
A blank semi-transparent black overlay will mask the UI


colbr (approved):
Thanks @SalvoAgustin good work

PR #156, opened 2021-09-03
Updated bid data to be updated automatically

SalvoAgustin:
WIP, changed to fetch with an interval of 5s, with a clear on useEffect clean. Task card:
https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/1e084c99-12dc-4489-8d0c-43f8bd39c91c

### DIFF
--- a/src/containers/NFTView/NFTView.tsx
+++ b/src/containers/NFTView/NFTView.tsx
@@ -172,7 +172,7 @@ const NFTView: React.FC<NFTViewProps> = ({ domain, onTransfer }) => {
 		if (
 			!isMounted.current ||
 			!znsDomain.domain ||
-			(znsDomain.domain && !znsDomain.domain.metadata && znsDomain.domain.image)
+			(!znsDomain.domain.metadata && znsDomain.domain.image)
 		)
 			return;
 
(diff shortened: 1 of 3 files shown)


colbr (changes requested):
In `lib/zAuction.ts:105`, we're caching bid data for 60 seconds. If you're updating the bid array every 5 seconds, you'll get the cached bids 12 times before you get the latest from the API. If we're wanting to update the bids every 5 seconds, there will need to skirt the cache in some instances.

One thing that would help with UX -- the user doesn't know the bids are automatically refreshing until they see a new bid appear. It would be good to have a countdown the user can see - something like "Refreshing bids in 15 seconds..." with a loading indicator. There is no design in place for this yet, I'll pass this onto the designers for them to work on while you change the first problem.

**Todo**
- Add ability to bypass the bid cache when we need to (it should **always** default to using the bid cache though)
- Bypass the bid cache in NFT View

SalvoAgustin:
Understood, I will put energy into that

colbr:
@SalvoAgustin just giving you an update on this one --

I think this feature may be made obsolete by the SDK. I'll have a look into it and let you know

colbr:
Closing for now - I believe this has been made obsolete by the SDK

PR #164, opened 2021-09-10
Cloudinary integration

colbr:
QA Instructions: https://www.loom.com/share/b6ceca159dd94e41b67aed1838c6eb7a

______

**Problem**

Loading full size images from IPFS is killing performance throughout the dapp. Currently, whenever the dapp renders an NFT image, it has to pull the full-size image and scale it down to whatever size it needs. This is horrible for bandwidth, and equally bad for performance.

Loading a single domain table with multiple NFT videos would cost the user 100s of MB of bandwidth, and would take a very long time to load. The extra kicker is that the files won't be cached since they're too big, so each page load would require redownloading each video. There are also no video posters, so the video takes a while to show anything.

Scaling down full-size images/videos is also resource intensive, and can create some awful screen tearing when scrolling. Using actual thumbnail sized images instead of scaled down high-res images is going to majorly improve the performance of the dapp.

**Solution**

We should use Cloudinary to generate different sized views of the original NFT, and render them in place. For example, we should use Cloudinary to generate thumbnail-sized versions of the original NFT for the domain table. This will **greatly** cut down on network and computational resources.

Problem & Solution Outline: https://www.loom.com/share/006179e42b0943a1bd00ff8a7343182e

___________________

**Progress**

_10.09.2021 8.45pm NZT_
Currently waiting for Cloudinary to be implemented in the "upload to IPFS" back-end flow - until that is done I won't have any way to thoroughly test the NFT Media component is working properly. It appears to be doing the right thing on the mainnet NFTs which is a good sign.

_12.09.2021 1pm NZT_
First QA request


stacy-hunt (approved):
Well that's a damn site faster than before! Nice work

colbr:
Thanks @stacy-hunt !

PR #167, opened 2021-09-12
Release 0.8.3

colbr:
Changes
1. Cloudinary integration (#164 --- #166 )

Cloudinary brings us massive improvements to performance & network usage. We now load Cloudinary images in place of IPFS media, with IPFS as a fallback. Cloudinary also generates many different sizes of each asset, so we can render thumbnails as actual 50x50 thumbnails rather than scaling down a large image.


stacy-hunt (approved):
Get it out there!

PR #170, opened 2021-09-16
Work/leo/update history with transfers

SalvoAgustin:
Created a hook to fetch with a polling interval the transfers, and added a method to fetch async the mint data of the nftId on provider

Left the comments so you can test


stacy-hunt (approved):
Working as expected

PR #171, opened 2021-09-16
Basic AIR WILD auction countdown

colbr:
Added a very basic temporary countdown for the end of the AIR WILD Season 0 auction


stacy-hunt (approved):
Tested, working as intended

PR #178, opened 2021-09-19
 Future Buttons Colours Updated

JeffersonMat:
Loom link: https://www.loom.com/share/4e64fe2be6014b36aec9572691849b5c 
Figma: https://www.figma.com/file/EO5VJm6sofH7rBWZzr8GLi/ZERO-Design-System?node-id=27%3A312

1. Make a Bid buttons colours have been updated
2. Added an un-active state to buttons. Make a bid state changes depending on if Wallet is disconnected
3. Colour in animation has been also updated and behaviour adjusted slightly.
4. View Button colour has also been updated
5. Colour of View Button has been updated when animation is running
6. Added an un-active state if button ever needs to be un-active
7. Animation behaviour has been also adjusted slightly

Let me know any questions.


src/components/Buttons/FutureButton/FutureButton.tsx
```diff
@@ -23,12 +24,14 @@ const FutureButton: React.FC<FutureButtonProps> = ({
 	glow,
 	loading,
 	alt,
+	isWalletConnected,
```
colbr, line 27: Noticed that this prop isn't being used anywhere. Make sure to prune out unused stuff.

Maybe a bit unrelated to this PR, but I noticed that this variable is acting like a boolean but it's typed as a string. Be careful to get your types correct - this wouldn't throw an error but it's confusing for other developers. You're passing the wallet ID from the parent, but you should be passing in a boolean with the value of `wallet ID !== undefined`. If a variable starts with "is" it should pretty much always be a boolean.

In terms of architecture, the buttons don't need any context for wallet state, etc. they just need to know if they should glow or not. Whether they should glow or not should be decided by the parent - this later is purely presentational. `glow` handles this functionality.

src/components/Buttons/FutureButton/FutureButton.tsx
```diff
@@ -39,16 +42,16 @@ const FutureButton: React.FC<FutureButtonProps> = ({
 
 	return (
 		<button
-			className={`${styles.futureButton} ${isSelected ? styles.selected : ''} ${
-				glow ? styles.glow : ''
-			} ${loading ? styles.Loading : ''} ${alt ? styles.Alt : ''} ${
-				className ? className : ''
-			}`}
+			className={`${styles.futureButton} ${
+				glow ? ' ' : styles.futureButtonUnActive
```
colbr, line 46: A couple minor things: 
1. I would use `''` instead of `' '` here because if you're adding a bunch of spaces to the class list it'll get quite unruly
2. "UnActive" is slightly incorrect grammatically, it should be "Inactive"

The above aren't showstoppers but would be good to fix

src/components/Buttons/FutureButton/FutureButton.tsx
```diff
@@ -39,16 +42,16 @@ const FutureButton: React.FC<FutureButtonProps> = ({
...
+			} ${isSelected ? styles.selected : ''} ${glow ? styles.glow : ''} ${
+				loading ? styles.Loading : ''
+			} ${alt ? styles.Alt : ''} ${className ? className : ' '}`}
 			onMouseEnter={handleHover}
 			onMouseUp={handleClick}
 			style={style}
 		>
-			<div className={styles.content}>
+			<div className={`${styles.Content}`}>
```
colbr, line 54: Minor: If it's just one class you don't need to wrap it like `'${styles.Content}'`, you can just do `className={styles.Content}`

src/components/Buttons/FutureButton/FutureButtonStyle.module.css
```diff
@@ -1,39 +1,63 @@
...
+		rgba(82, 87, 255, 0.25) 0%,
+		rgba(82, 203, 255, 0.25) 100%
+	);
 	--box-shadow: rgba(0, 209, 255, 0.1);
 	--box-shadow-heavy: rgba(0, 209, 255, 0.5);
+	--border-color: #3ca1ff;
+	--border-color-hover: #fff;
+}
+
+.futureButtonUnActive {
```
colbr, line 14: Minor: same deal w UnActive vs Inactive - not a huge issue

colbr (changes requested):
Couple tiny code style changes, but overall good work

src/components/Buttons/FutureButton/FutureButton.tsx
JeffersonMat: Yeah, I actually know this, I think I just missed it.

src/components/Buttons/FutureButton/FutureButtonStyle.module.css
JeffersonMat: Sweet, grammar have been changed and notes were taken.

colbr (changes requested):
Messaged you about this one in zer0 last week but I'll put it in here for book-keeping --

Getting a blue glow when hovering over disabled buttons
Screen Shot 2021-10-12 at 5 07 00 PM

colbr (approved):

PR #179, opened 2021-09-21
Bids History padding and glow bleeding zNAlink

JeffersonMat:
Bleeding Glowing:

Fixed the width of the glowing so it doesn't bleed into the text. –ZNA link.

The fix works regardless of the size of the text.

Bid History:

1. -Previously, as it scrolled down, the list got cut off by the edge of the box (at the bottom). I've added a padding bottom so it looks cleaner and more consistent.
 
2. -A padding has also been added to the loader state while it loads the list so it's centered height-wise.

3. -Finally, padding was also added when it show "No bids". 

Let me know any questions.

Link Bids History: https://www.loom.com/share/c08d2aef408f46b9a164375d5f70a9a4
Link to glowing bleeding: https://www.loom.com/share/923674f6b33e4c1f9bc4187caf027b2c


src/containers/NFTView/NFTView.tsx
```diff
@@ -231,7 +231,9 @@ const NFTView: React.FC<NFTViewProps> = ({ domain, onTransfer }) => {
 
 	const history = () => (
 		<section
-			className={`${styles.History} ${styles.Box} blur border-primary border-rounded`}
+			className={`${styles.History} ${
+				bids && bids.length === 0 ? styles.HistoryNoBids : ' '
```
colbr, line 235: You can write the conditional here as `bids?.length === 0 ?`, saves a lil bit of space. Obviously not needed for approval, just letting you know that's a thing. It's called [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

colbr (approved):
Nice, good stuff @JeffersonMat

src/containers/NFTView/NFTView.tsx
JeffersonMat: Cool. Sometimes I do what you've suggested but typescript gets funny about it sometimes. Thank you.

src/components/ZNALink/ZNALink.module.css
```diff
@@ -44,8 +44,8 @@
 	left: 50%;
 	-webkit-transform: translate(-50%);
 	transform: translate(-50%);
-	width: calc(100% - 16px);
-	box-shadow: 0 0 20px 13px #3ca1ff;
+	width: calc(100% - 28px);
```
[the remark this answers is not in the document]
JeffersonMat: Yeah, that was a hard one. I had to really zoom in but it gets all pixeled it so you can't really see it clearly.

PR #180, opened 2021-09-21
Switch Fathom from Auto to Hash

akshatmittal:
Switches Fathom Integration to force hash based routing instead of enabling auto-detection. This will allow us to see visits and users for specific namespaces on Fathom.


colbr (approved):
Thanks for sorting that out

PR #183, opened 2021-09-25
Better auction countdown banner

colbr:
Set up a nicer auction countdown timer. Still not the ideal solution we're after, but a much nicer iteration.

**Changes**
Added pre and post messaging - "auction starting in x", "auction ended". You can set an amount of time for these messages to be visible for, so auction start countdown starts 12 hours before the auction, and the auction end message disappears after 3 hours.

Set up values for next drop - may need changing


EthFish (approved):
Works as intended

PR #188, opened 2021-09-27
Helper debug logging when accepting bid

Remscar:


colbr:
@Remscar is this one still needed, or was it just for debugging?

PR #189, opened 2021-09-27
Client-side enforcement of lowercase domain names in Mint flow

colbr:
All domain names need to be lower case. This PR adds validation to prevent transactions happening if someone inputs a lowercase domain in the Mint flow.

Validation:
1. In the Information section of the Mint flow (i.e. page 1) - you can't continue with the flow if the domain has an uppercase character.
2. A check in the final submit method of the Mint flow, just in case the first bit of validation failed, or in case someone is trying to circumvent the first step.
3. An extra check in the Mint & Staking providers - this one should never be hit, but probably wise to have it in there.

Note:
In the requirements, it said uppercase characters should be automatically converted to lowercase as they're typed out, but that made the input feel broken. Instead, the dapp renders an error label below the input whenever the domain is invalid.

### DIFF
--- a/src/components/Inputs/TextInput/TextInput.tsx
+++ b/src/components/Inputs/TextInput/TextInput.tsx
@@ -27,6 +27,8 @@ type TextInputProps = {
 	resizable?: boolean;
 	alphanumeric?: boolean; // If we want only alphanumeric characters
 	numeric?: boolean;
+	lowercase?: boolean;
+	maxLength?: number;
 };
 
 const TextInput: React.FC<TextInputProps> = ({
@@ -42,6 +44,8 @@ const TextInput: React.FC<TextInputProps> = ({
 	resizable,
 	alphanumeric,
 	numeric,
+	lowercase,
+	maxLength,
 }) => {
 	//////////////////
 	// State & Data //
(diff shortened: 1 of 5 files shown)


src/lib/providers/MintProvider.tsx
```diff
@@ -56,6 +56,11 @@ const MintProvider: React.FC<MintProviderType> = ({ children }) => {
 	// }, []);
 
 	const mint = async (nft: NftParams, setStatus: (status: string) => void) => {
+		// @todo better validation
+		if (/[A-Z]/.test(nft.zna)) {
+			return;
```
Remscar, line 61: Nit: Maybe throw an error?

Remscar (approved):

src/components/Inputs/TextInput/TextInput.tsx
```diff
@@ -55,10 +56,20 @@ const TextInput: React.FC<TextInputProps> = ({
...
 		const newValue = event.target.value;
-		if (validate(newValue) && onChange) return onChange(event.target.value);
+		if (validate(newValue) && onChange) {
+			return onChange(format(event.target.value));
+		}
+	};
+
+	const format = (str: string) => {
+		if (lowercase) {
+			return str.toLowerCase();
```
Remscar, line 66: This doesn't seem to happen when requesting a domain to mint; ie; automatically changing capital letters to lowercase.

colbr: Currently the 'lowercase' prop isn't set on the input in the Mint flow -- I implemented this functionality, but it made the input feel broken so I opted for an error message instead. Left it in just in case we need it in future design iterations (probably will)

src/lib/providers/MintProvider.tsx
colbr: Good call - will do

EthFish (approved):
All works well. Good UX choice with error message behaviour

PR #191, opened 2021-09-28
Mint Wheels Banner

alejoviola:


src/components/Banners/MintWheelsBanner/MintWheelsBanner.tsx
```diff
@@ -0,0 +1,44 @@
+//- React Imports
+import React from 'react';
+
+//- Style Imports
+import styles from './MintWheelsBanner.module.css';
+
+type MintWheelsBannerProps = {
+	title: string;
+	label: string;
+	labelBold: React.ReactNode;
```
colbr: Make this whole label one ReactNode so we can send in any arbitrary components (countdown timer, etc.)

colbr (changes requested):
Awesome work Alejo, just one small change and we're good

colbr (approved):

PR #195, opened 2021-09-30
[FEATURE] Bid button behaviour when wallet not connected

padala:
**Summary**: Users are confused as to why they cannot make bids. This is because they do not have a wallet connected and there is no messaging to inform the user of what went wrong. We added a new prompt to inform users that wallet is not connected.

**Task URL**: https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/a1cfd46a-9748-4aca-9bff-133c869dd49f


colbr (approved):
Excellent job, very clean code! Thanks @padala

PR #197, opened 2021-10-01
Default use Infura for web3 RPC connections.

Remscar:


colbr (approved):
Awesome - looks good!

PR #213, opened 2021-10-07
[TWEAK] Switch notification system to Redux

ericlowturing:
Summary:
1) Setup the redux store for the whole application
2) Added the notifications store inside the global redux store. We can add more stores from this.
3) Removed Notifications context provider and updated the hooks to call notifications redux store values and redux actions.
4) Updated the NotificationDrawer component to use redux store instead of context.
5) Added unit tests for notifications redux store, useNotifications hooks and NotificationDrawer component (I recommend to add unitests for added/updated components/hooks from now since we have unitests stuffs)

Task:
https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/a7b0fb76-0368-4c14-b662-6c8d91737d7d


padala:
@ericlowturing @colbr  I'll check this out today. Thanks

src/store/notifications/reducer.ts
```diff
@@ -0,0 +1,45 @@
...
+export const REDUCER_NAME = 'notifications';
+
+export const INITIAL_STATE: NotificationsState = {
+	notifications: [],
+};
+
+const reducer = (state = INITIAL_STATE, action: NotificationsActions) => {
+	switch (action.type) {
+		case ADD_NOTIFICATION_REQUEST:
+			const { id, text, duration } = action.payload;
```
padala, line 20: Can we add the types here itself?

src/store/reducers.ts
```diff
@@ -0,0 +1,18 @@
+import { combineReducers } from 'redux';
+import { connectRouter } from 'connected-react-router';
```
padala, line 2: Do we need redux for routing? Can we do it in a different PR?

src/store/reducers.ts
```diff
@@ -0,0 +1,18 @@
...
+import notificationsReducer, {
+	REDUCER_NAME as NOTIFICCATION_REDUCER_NAME,
+} from './notifications/reducer';
+
+const createRootReducer = (history: History<any>) =>
+	combineReducers({
+		router: connectRouter(history),
+		[NOTIFICCATION_REDUCER_NAME]: notificationsReducer,
+		/**
+		 * Other reduceers will be added here
```
padala, line 14: minor typo here

src/store/notifications/actionTypes.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
padala, line 1: Can we add an `index.ts` barrel file at `store/notifications`?

src/components/NotificationDrawer/NotificationDrawer.tsx
```diff
@@ -1,41 +1,42 @@
...
 //- Style Imports
 import styles from './NotificationDrawer.module.css';
 
-const NotificationDrawer = () => {
-	// TODO: Need to animate notifications coming in and out
-	// TODO: Change to useMemo
-	// TODO: Change notifications over to a portal?
-
-	const [visibleNotifications, setVisibleNotifications] = useState<any>([]);
+export const TEST_ID = {
```
padala, line 11: Can we rename this to `TEST_ID_ATTRIBUTES` ?

padala (changes requested):
@ericlowturing  Added minor comments. Let me know if you want to connect to discuss any of the items. Really liked the detailed implementation.
Thanks :)

src/store/reducers.ts
ericlowturing: We need it to access routing properties in the component. It is super light redux for the router. I think it is no problem to keep it.

src/store/reducers.ts
ericlowturing: Fixed it

PR #215, opened 2021-10-09
Updated history transfers with "from" and "to"

SalvoAgustin:
Updated history with "from" and "to" parameters of transfers, fetched directly from subgraph


SalvoAgustin:
@colbr Yes, as long as you want to merge the sdk first, the other that make the calls directly to the subgraph can be closed

PR #227, opened 2021-10-16
Show USD price alongside WILD price in tables

colbr:


EthFish (changes requested):
Change text color to grey 3 as per [Figma](https://www.figma.com/file/0zFJaRNODss8lxwRygUh24/UI-Wilder-Market?node-id=6039%3A344654)

colbr:
> Change text color to grey 3 as per [Figma](https://www.figma.com/file/0zFJaRNODss8lxwRygUh24/UI-Wilder-Market?node-id=6039%3A344654)

Done

PR #251, opened 2021-10-22
Fix missing bids in NFT View History

colbr:
**Bug**

As per images below, some bids were missing from the bid history but were still being calculated in the "Highest Bid" calculation. I fixed this by removing all zAuction API calls and only grabbing NFT event data through the SDK.

**Bug Report**

rnjfks0jtcfoddeixb4s
srzv5ruyplqopnliqmbw
ot9i8hstxxlimfizb1jz


EthFish (approved):
Looks good. Tested by; 
1. bidding on kovan and checking top bid in list view, and in domain history
2. Checking some mainnet domains to see if top bid in nft history was shown in list

Works as expected

PR #281, opened 2021-10-29
feat(nft-view): fixed margins for nft view

padala:


colbr (approved):
Looks great, thanks @padala

PR #290, opened 2021-11-02
Share to twitter

padala:
**Note**: I updated the netlify environment variables with the current zns share url.


colbr (approved):
Looks great!

colbr:
Holding off on merging this one until we have the the API is:
- Deployed on the correct Heroku account
- Given a domain name
- Migrated to the zer0-os git account

PR #293, opened 2021-11-04
Disabled pinch zooming with two fingers.

JeffersonMat:
I've added a line to the header meta tag, to disable user scalability. 
And also, added touch-action to CSS to only enable x and y action for scrolling.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -190,7 +190,7 @@ html,
 body {
 	height: 100%;
 	overflow: hidden;
-    touch-action: pan-x pan-y;
+	touch-action: pan-x pan-y;
 }
 
 #main-container {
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -190,6 +190,7 @@ html,
 body {
 	height: 100%;
 	overflow: hidden;
+    touch-action: pan-x pan-y;
```
colbr, line 193: Make sure to fix indentation. I'm not sure how this wasn't caught by Prettier though, can you double check Prettier is running on save?

colbr (approved):
Nice @JeffersonMat 

Requested a fix for some indentation, but I won't let that block the merge. Push it through with proper indentation then merge :)

src/styles/main.scss
JeffersonMat: Done. I have run npm, however, I don't seem to see prettier running on commits. I'll dig into this to find out what's going on.

JeffersonMat:
The indentation issue has been fixed.

PR #295, opened 2021-11-04
build: enforce conventional commit messages

colbr:
Zer0 Card: https://zer0.io/a/network/tasks/board/e85a99cf-665f-427c-bb0b-2b4e12ba207a/90d27375-c9ff-4b04-8391-bbe856249c16

--------

**Problem**
We want to enable automatic versioning, but to do so we need everyone to be writing [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716). We need a way to enforce all commit messages follow these conventions.

**Solution**
- Install `commitlint`
- Run `commitlint` on commit
- Throw an error if commit message breaks any rules
- Let all devs know of changes


colbr:
Going to create a task to go through and fix all the error types, as was changed in 02116b0

colbr:
Jury is still out on whether this one is a good idea or not, stalling it for now

colbr:
Decided this wasn't needed for now

PR #311, opened 2021-11-30
Update Wheels flow to work with Cribs contract

colbr:
@Remscar made some changes to the existing sale flow:
- Added helpers for protocol change
- Switched contract address from the wheels sale contract to the cribs sale contract


colbr:
@Remscar wrote the code, I made the PR and approved it

Does not need design QA as we're just merging into another feature branch

PR #335, opened 2021-12-20
Wheels Launch for 21st Dec 2021

akshatmittal:
Wheels Launch Sale for 21st Dec 2021.


stacy-hunt (approved):
Mint flow, timer and banner copy tested and working properly

PR #336, opened 2021-12-21
Use default wallet connector

colbr:
dApp now defaults to using Infura as wallet connector when user wallet is not connected. This means we can show TVL and APR without prompting the user to connect.


colbr:
@stacy-hunt this is the one - hit Approve if you think it's good to go. 

If I'm not around, you can also just hit "Squash and merge" - might have to click the arrow beside the 'Create a merge commit' button to see it. This will push the changes to production 😊

stacy-hunt (approved):
Default Wallet is working now

PR #342, opened 2021-12-25
App Switcher & Staking + Market Integration

akshatmittal:


colbr:
This is looking great @akshatmittal - I can't speak on the design (lgtm), but the one major thing we're missing in terms of functionality is the ability to change Nav content based on the current page/dapp. I'm sure you're already well aware, but on the Market dapp we show a zNA on the left and an array of market related buttons on the right, whereas on the Staking dapp we show a page title on the left and not much else on the right. It would be great if on each page we could call some sort of `setNav` method to provide component(s) for the Nav. This could be implemented through a simple hook/provider.

Awesome work, things seem to be working very smoothly!

akshatmittal:
@colbr Implemented exactly that! Also added the fix for the wallet connection differences between Staking and Market.

colbr (changes requested):
Looking great, nice work @akshatmittal 

One small thing which must have been missed when merging --
The search bar has reappeared above the deposits table - there is no search bar in the design. Should be easy enough to remove, you should be able to set a prop in `DepositTable.tsx` on the `GenericTable` component.

Screen Shot 2022-01-13 at 10 56 27 AM

akshatmittal:
@colbr Done! I definitely missed that!

colbr (commented):
Looking great @akshatmittal 

Final piece of the puzzle is what to do with those back/forward buttons within the staking app. I think we're teeing up a discussion regarding those in the next couple days, perhaps a temporary solution could be to hide them on staking since they're not linked up to anything.

colbr (approved):
Looking great, ready to merge imo!

PR #345, opened 2022-01-04
Add flexible formatByDecimalPlace util function and tests

domw30:
Link to associated zer0 card:  [Notion Card](https://www.notion.so/rational-nomads/67a5d1b61d4146f491d5996a17b14a81?v=4f04ed31a23340bfbbdc5560502bc429&p=4d62ccf715fc40e4a0ca203edea1ced1)

-----

## 1. Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] zer0 card has been moved to the Code Review column
- [x] zer0 card has a link to this PR
- [x] A reviewer has been tagged in the zer0 card

## 2. PR type
- [x] Feature

## 3. What is the current behaviour?
Existing format number util doesn't allow us to pass a desired amount of decimal places to display in the UI on a value.

New formatByDecimalPlace util allows us to add a desired amount of decimal places to display. Fully tested.

## 5. Other information
The reason for this ticket was due to displaying wild percentage change in ui. This PR will need to be merged first.
[Loom](https://www.loom.com/share/65afb1e2c2a1429096a6ff65d5535cd5)

### DIFF
--- a/src/lib/utils/number.ts
+++ b/src/lib/utils/number.ts
@@ -13,10 +13,6 @@ export const formatEthers = (number: string) => {
 	return formatNumber(asNumber);
 };
 
-export const ONE_DECIMAL_PLACE_VALUE = 1;
-export const TWO_DECIMAL_PLACE_VALUE = 2;
-export const THREE_DECIMAL_PLACE_VALUE = 3;
-
 export const formatByDecimalPlace = (
 	value: number | string,
 	places: number,
--- a/src/lib/utils/number.ts
+++ b/src/lib/utils/number.ts
@@ -12,3 +12,17 @@ export const formatEthers = (number: string) => {
 	const asNumber = Number(ethers.utils.formatEther(number));
 	return formatNumber(asNumber);
 };
+
+export const ONE_DECIMAL_PLACE_VALUE = 1;
+export const TWO_DECIMAL_PLACE_VALUE = 2;
+export const THREE_DECIMAL_PLACE_VALUE = 3;
+
+export const formatByDecimalPlace = (
+	value: number | string,
+	places: number,
+	localeCode?: string,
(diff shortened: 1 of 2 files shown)


domw30:
@padala / @colbr - as per Stacy's request are you able to review this PR please, should be straight forward. I'd like to merge this into develop branch and pull updates into the show 24hr token price percentage change task which only requires this formatting to wrap it up too. This PR will require merging first. Thanks !

src/lib/utils/number.ts
```diff
@@ -12,3 +12,17 @@ export const formatEthers = (number: string) => {
 	const asNumber = Number(ethers.utils.formatEther(number));
 	return formatNumber(asNumber);
 };
+
+export const ONE_DECIMAL_PLACE_VALUE = 1;
+export const TWO_DECIMAL_PLACE_VALUE = 2;
+export const THREE_DECIMAL_PLACE_VALUE = 3;
```
padala, line 18: I think we can remove these const and just pass the number unless we want to restrict the decimal places just to 3.

src/lib/utils/number.ts
```diff
@@ -12,3 +12,17 @@ export const formatEthers = (number: string) => {
...
+
+export const formatByDecimalPlace = (
+	value: number | string,
+	places: number,
```
padala, line 22: If we want to restrict as mentioned above then the type needs to be modified from number

padala (changes requested):

src/lib/utils/number.ts
domw30: sounds good to me, I'll remove these and update tests

src/lib/utils/number.ts
domw30: as comment above

padala (approved):

PR #351, opened 2022-01-19
Release 0.2.0

colbr:
**Changes:**
- Added a "Total Staked USD" widget to the Deposits page - [Associated Notion Card](https://www.notion.so/Add-a-widget-to-Deposits-to-show-user-s-total-value-locked-in-USD-21112f728e894f04aa9ed725b3648bb9)
- Fixed a bug where dApp would crash on My Deposits page - [Associated Notion Card](https://www.notion.so/Fix-My-Deposits-page-crash-138ab13b61ae4c0ca5c70dea106b7844)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
 				"@web3-react/walletconnect-connector": "^6.2.0",
 				"@web3-react/walletlink-connector": "^6.1.9",
 				"@zero-tech/zauction-sdk": "0.0.30",
-				"@zero-tech/zfi-sdk": "^0.1.12",
+				"@zero-tech/zfi-sdk": "^0.1.13",
 				"@zero-tech/zns-sdk": "0.0.35",
 				"apollo": "^2.32.5",
 				"autosize": "^5.0.1",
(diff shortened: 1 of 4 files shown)


EthFish (changes requested):
Widget slow but working fine
Switching to My Deposits does not crash app
Small bug encountered when switching tabs, see [this loom video](https://www.loom.com/share/3ac551e7be6341b2936fe13748eb9d92)

colbr:
> Widget slow but working fine

Agreed, it is quite slow. I think particularly for the wallet we're using, because on Kovan it has like 30 deposits to scan through. When I try on my real wallet with only 2 deposits, it works very fast.

> Small bug encountered when switching tabs

This is because when the window becomes scrollable, the scroll bar width pushes the content to the side. It's happening in the preview build because you're on Kovan and have lots of deposits, so the window is scrollable. Need to sort that out in the entire dApp including zNS so will push this through and sort another day.

EthFish (approved):
OK, all sounds good!

PR #352, opened 2022-01-19
feat(kicks-drop): update the nft media to use animation_url if available

padala:

-----

## 1. Pull request checklist

### DIFF
--- a/src/components/Artwork/Artwork.tsx
+++ b/src/components/Artwork/Artwork.tsx
@@ -88,7 +88,12 @@ const Artwork: React.FC<ArtworkProps> = ({
 				size="tiny"
 				className={`${styles.Image} border-rounded`}
 				alt="NFT Preview"
-				ipfsUrl={metadata?.image_full || metadata?.image || ''}
+				ipfsUrl={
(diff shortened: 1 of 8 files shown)


src/components/Artwork/Artwork.tsx
```diff
@@ -88,7 +88,12 @@ const Artwork: React.FC<ArtworkProps> = ({
 				size="tiny"
 				className={`${styles.Image} border-rounded`}
 				alt="NFT Preview"
-				ipfsUrl={metadata?.image_full || metadata?.image || ''}
+				ipfsUrl={
+					metadata?.animation_url ||
```
colbr, line 92: Artwork shouldn't use `animation_url`. Artwork components are just little thumbnails. I think we decided a while ago to just leave them as images. `NFTMedia` components with `size="tiny"` will just use a thumbnail **or** the first frame of the video - it ends up looking nicer if we just use the OpenSea thumbnail i.e. `image`.

colbr (changes requested):
Looks good, one small change

colbr (approved):

PR #371, opened 2022-02-02
TextButton test created. Converted file tsx. 

JeffersonMat:
[Associated Notion Card](https://www.notion.so/rational-nomads/Tests-for-Text-Button-component-2e7825cc3e5b4a9eadfe16f9a81db8a4)

-----

## 1. Pull request checklist

- [X] Notion card has been moved to the Code Review column
- [X] Notion card has a link to this PR
- [X] A reviewer has been assigned to the Notion card

## 3. What is the old behavior?

TextButton component had js. extension and had no test for it.

Converted component to .tsx extension. Added Type for incoming props. Created a test file for the component.


colbr (approved):
Looking good. I noticed there isn't a test for the `toggleable` functionality, but I don't think it is actually being used anywhere so probably worth taking it out of the component completely. We can do that later though, for now it's looking good 😊

PR #373, opened 2022-02-02
Fix eslint warnnings

ericlowturing:

-----

## 1. Pull request checklist

### DIFF
--- a/src/containers/other/NFTView/NFTView.tsx
+++ b/src/containers/other/NFTView/NFTView.tsx
@@ -26,8 +26,6 @@ import useCurrency from 'lib/hooks/useCurrency';
 import { toFiat } from 'lib/currency';
 import { chainIdToNetworkType, getEtherscanUri } from 'lib/network';
 import { useZnsContracts } from 'lib/contracts';
-import { getDomainId } from 'lib/utils';
-import { useZnsDomain } from 'lib/hooks/useZnsDomain';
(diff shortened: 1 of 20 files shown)


src/containers/other/NFTView/NFTView.tsx
```diff
@@ -411,10 +409,11 @@ const NFTView: React.FC<NFTViewProps> = ({ onTransfer }) => {
 		return (
 			<>
 				<div className={styles.Stats}>
-					{data.map((item) => (
+					{data.map((item, index) => (
 						<>
 							{!item.isHidden ? (
 								<StatsWidget
+									key={index}
```
colbr, line 416: Use something other `index` as the `key` here - [iirc it's bad practice to use index](https://stackoverflow.com/questions/46735483/error-do-not-use-array-index-in-keys). I don't think we would run into problems (currently) by using `index` here, but best to change it in my opinion.

colbr (changes requested):
One small change to make. Other than that, looking good! Thanks @ericlowturing

src/containers/other/NFTView/NFTView.tsx
ericlowturing: Agreed! Just updated the key so that we won't have redundant key value. Thanks

colbr (approved):
Nice!

PR #374, opened 2022-02-07
Release 0.11.3 - MIME type hotfix

colbr:
NFTMedia was previously relying on IPFS to grab MIME type. IPFS is less reliable than Cloudinary, so we should use Cloudinary as our source of MIME type.

Further adjustments to logic will be needed, but this works as a bandaid.

### DIFF
--- a/src/components/NFTMedia/index.tsx
+++ b/src/components/NFTMedia/index.tsx
@@ -23,6 +23,7 @@ import CloudinaryMedia from './CloudinaryMedia';
 // Library Imports
 import classNames from 'classnames/bind';
 import { getHashFromIPFSUrl } from 'lib/ipfs';
+import { generateCloudinaryUrl } from './config';
 
 // Possible media types based on
 // MIME type of content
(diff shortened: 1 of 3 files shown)


src/components/NFTMedia/index.tsx
```diff
@@ -38,22 +39,16 @@ const cx = classNames.bind(styles);
...
-	return new Promise((resolve, reject) => {
-		fetch('https://gateway.ipfs.io/ipfs/' + hash, { method: 'HEAD' })
-			.then((r: Response) => {
-				const contentTypeHeader = r.headers.get('Content-Type');
-
-				if (contentTypeHeader?.startsWith('image')) {
-					resolve(MediaType.Image);
-				} else if (contentTypeHeader?.startsWith('video')) {
+	return new Promise((resolve) => {
+		fetch(generateCloudinaryUrl(hash, 'video'), { method: 'HEAD' }).then(
```
padala, line 43: @colbr  Do we intend to check cloudinary instead of ipfs and see if the video is available or not?

colbr: @padala yep, IPFS has been really patchy, so I think we should check Cloudinary instead. Going to do an extra iteration on this one, this PR is just a hotfix for the videos that aren't loading

padala (approved):

src/components/NFTMedia/index.tsx
padala: @colbr yeah ipfs has been patchy.

PR #377, opened 2022-02-07
[BUG] - Hide ‘Mint’ button on NFT page that user owns

domw30:
[Associated Notion Card](https://www.notion.so/rational-nomads/67a5d1b61d4146f491d5996a17b14a81?v=4f04ed31a23340bfbbdc5560502bc429&p=e2da8a09a46944a080d0a064eda63c71)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Bug Fix

All instances of the MINT button were previously removed from the Title Bar due to confusion for users when minting wheels/kicks. However, the MINT button was still visible when a user was logged in and on one of their owned NFT domains. 

https://user-images.githubusercontent.com/39112648/152789227-5bb0d06b-7e00-457b-a6b0-9a3cc48288cd.mov

MINT button is no longer visible to the user.

## 5. Other information
I have created a ```featureFlag``` util so we can use this in future to display/hide features that we want to see in ```development``` but do not want in ```production``` environment rather than removing/commenting out code.


colbr (changes requested):
Hey @domw30 -

Can we add a way for anyone with authority to mint to be able to see the Mint button? Some of the Wilders will be wanting to mint NFTs, and I believe they use the UI for it. It would be ideal if they could run a console command to make it re-appear

padala:
> Hey @domw30 -
> 
> Can we add a way for anyone with authority to mint to be able to see the Mint button? Some of the Wilders will be wanting to mint NFTs, and I believe they use the UI for it. It would be ideal if they could run a console command to make it re-appear

@colbr @domw30 I think we can add a query param like `?betaMode=true` which when added to the URL can enable these kinds of beta features.

padala:
@colbr @domw30  On a different note I like the feature flags concept but not sure if it makes sense to use it for hiding mint nft as we want to hide it all together irrespective of the environments. Not sure if there is any change. Thanks

domw30:
> > Hey @domw30 -
> > Can we add a way for anyone with authority to mint to be able to see the Mint button? Some of the Wilders will be wanting to mint NFTs, and I believe they use the UI for it. It would be ideal if they could run a console command to make it re-appear
> 
> @colbr @domw30 I think we can add a query param like `?betaMode=true` which when added to the URL can enable these kinds of beta features.

Thanks both. I didn't realise there were some users that would be using the ```Mint``` flow in the production environment. I would assume that these users are developers/designers/QA/testers etc, no probs, sounds good to me!

I've been looking at how best to run a console command to toggle the beta features, @colbr do you mean running a command in the dev tools?

shall take a look at query string option too. Query string is a nice option and gives quick access to beta features, but am I right in saying there would be no way to restrict a public user visiting the beta url?

colbr:
> am I right in saying there would be no way to restrict a public user visiting the beta url

Yep, this will be the same for any method we implement purely in the front end. The minting contract is currently only allowing a few wallets to mint, so if someone wants to manipulate the UI to try and mint the contract call will just fail anyway.

>  @colbr do you mean running a command in the dev tools

Yup! Previously we had a command you could run in dev tools that showed what future iterations of the dapp would look like. Check out [MvpVersionProvider](https://github.com/zer0-os/zNS-dapp/blob/develop/src/lib/providers/MvpVersionProvider.tsx) for an example of this. For context, the reason we're after a quick (& potentially not ideal) solution is because we're re-adding the mint button back once advanced metadata is properly implemented.

domw30:
Hey @padala, @colbr, ready for another review - I think notion notifications may have been missed/buried under others. Loom video on task card explains the latest changes - using ```mvpVersion```

src/lib/utils/featureFlags.ts
```diff
@@ -0,0 +1,33 @@
+export enum Environments {
+	TEST = 'test',
+	DEVELOPMENT = 'development',
+	PRODUCTION = 'production',
+}
+
+const isEnvironmentTest = (environment: string) =>
+	environment === Environments.TEST;
```
padala, line 8: @domw30  Can we get rid of this file as we're no longer using it?

domw30: Yeah sure @padala, I will remove now and ping you once done

padala: Probably associated tests too.

padala (approved):

src/lib/utils/featureFlags.ts
padala: @domw30  Approved the PR but I think @colbr  should also approve it as he requested changes earlier on this.
Thanks

domw30:
Not 100% sure why the build fails here. Test suite running okay locally. I'm restricted from re-running the build.
Screenshot 2022-02-21 at 15 29 17

colbr (approved):
Nice one mate

PR #399, opened 2022-02-28
Circle CI build failing fix

domw30:
[Associated Notion Card](https://www.notion.so/dApp-Core-Product-Team-Board-e96437225e264ae8ae8d46ca5f4d7b35?p=585150edf7844452bfffc96f40737e87)

-----

## 1. Pull request checklist

- [x] Notion card has been moved to the Code Review column
- [x] Notion card has a link to this PR
- [x] A reviewer has been assigned to the Notion card

## 2. PR type
BugFix

## 3. What is the old behaviour?
Circle Ci build fails on Preview Card Test
Screenshot 2022-02-28 at 12 39 26

Screenshot 2022-02-28 at 12 39 44

```A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks```

## 4. What is the new behaviour?
Circle Ci build passes

Added detectOpenHandles
Attempt to collect and print open handles preventing Jest from exiting cleanly. Use this in cases where you need to use --forceExit in order for Jest to exit to potentially track down the reason. This implies --runInBand, making tests run serially. Implemented using [async_hooks](https://nodejs.org/api/async_hooks.html).

Added maxWorkers
Alias: -w. Specifies the maximum number of workers the worker-pool will spawn for running tests. In single run mode, this defaults to the number of the cores available on your machine minus one for the main thread. In watch mode, this defaults to half of the available cores on your machine to ensure Jest is unobtrusive and does not grind your machine to a halt. It may be useful to adjust this in resource limited environments like CIs but the defaults should be adequate for most use-cases.

Screenshot 2022-02-28 at 13 13 16

PR #405, opened 2022-03-03
feat: Cancel bid

colbr:
[Associated Notion Card](https://www.notion.so/Ability-To-Cancel-Bids-584d100765054812921c8447b2148ca9)

-----

## 1. Pull request checklist

## 2. PR type
 

Feature

There was no way to see active bids, and there was no way to cancel bids.

There is now a bid table in the user profile, and a flow for cancelling an individual bid.

## 5. Other information

Not quite ready to be tested!

There's a failed test in the CI, it's for `PreviewCard.tsx` - this was fixed in #399


colbr:
@ericlowturing a couple things missing in this, but could you let me know how it's looking so far? Thanks 😊

p.s. for testing I just wrote one test file which tests both the hooks and the component

src/components/Profile/Profile.tsx
```diff
@@ -88,13 +88,11 @@ const Profile: React.FC<ProfileProps> = ({ id, yours, onNavigate }) => {
 					<CopyInput value={id} />
 				</div>
 			</div>
-			<TabBar tabs={['Your Domains']} onSelect={select} />
+			<TabBar tabs={['Your Domains', 'Your Bids']} onSelect={select} />
```
ericlowturing, line 91: nit: I think we can use it with constants:

```
const enum TABS = {
  DOMAINS = "Your Domains",
  BIDS = "Your Bids"
};
```

```
tabs = {[TABS.DOMAINS, TABS.BIDS]}
```

src/components/Profile/Profile.tsx
```diff
@@ -88,13 +88,11 @@ const Profile: React.FC<ProfileProps> = ({ id, yours, onNavigate }) => {
...
 			{selected === 'Your Domains' && (
 				<OwnedDomainsTable onNavigate={navigateToDomain} />
 			)}
-			{/* {selected === 'Offers' && (
-				<RequestTable onNavigate={navigateToDomain} userId={id} />
-			)} */}
+			{selected === 'Your Bids' && <BidTable />}
```
ericlowturing, line 95: And we can use here `{selected === TABS.BIDS && <BidTable />}`

src/components/Wizard/Presets/NFTDetails.tsx
```diff
@@ -39,8 +39,12 @@ const NFTDetails = ({
 				<h2>0://{domain}</h2>
 			</div>
 			{creator && <Member id={creator} subtext={'Creator'} />}
-			{otherDetails?.map((detail: Detail) => (
-				<DetailComponent text={detail.value} subtext={detail.name} />
+			{otherDetails?.map((detail: Detail, i) => (
+				<DetailComponent
+					key={i + detail.value}
```
ericlowturing, line 44: Nice to have a `key` here

src/containers/Tables/BidTable/BidTableRow.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const renderComponent = ({ data = mockData } = {}) => {
+	return render(
+		<table>
+			<tbody>
+				<BidTableRow data={data} />
+			</tbody>
+		</table>,
+	);
+};
```
ericlowturing, line 30: I think we can just use `return <BidTableRow data={data} />;` instead of using `<table>, <tbody>` tag since it is the test file of `BidTableRow` component.

src/containers/Tables/BidTable/BidTableRow.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+		const { getByTestId } = renderComponent();
+		const container = getByTestId(TEST_ID.CONTAINER);
+		expect(container.childElementCount).toBe(HEADERS.length);
+	});
+
+	it('should render bid amounts correctly', () => {
+		const { getByTestId } = renderComponent();
+		const yourBid = getByTestId(TEST_ID.YOUR_BID);
+		const highsetBid = getByTestId(TEST_ID.HIGHEST_BID);
+		expect(yourBid.textContent).toBe('500.0 WILD');
```
ericlowturing, line 49: nit: it is the same with `expect(yourBid).toHaveTextContent('500.0 WILD');`

src/containers/Tables/BidTable/BidTableRow.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+			data: {
+				...mockData,
+				highestBid: ethers.utils.parseEther('200'),
+				yourBid: ethers.utils.parseEther('100'),
+			},
+		});
+		const status = getByTestId(TEST_ID.STATUS);
+		expect(status.textContent).toBe('Outbid');
+	});
+});
```
ericlowturing, line 82: Looks great the tests

src/containers/Tables/BidTable/BidTableRow.tsx
```diff
@@ -0,0 +1,70 @@
...
+	domainName: string;
+	auctionId: string;
+	domainId: string;
+	domainMetadataUrl: string;
+	date: Date;
+	yourBid: ethers.BigNumber;
+	highestBid: ethers.BigNumber;
+};
+
+export const TEST_ID = {
```
ericlowturing, line 17: Nice! 👍

src/containers/Tables/BidTable/BidTableRow.tsx
```diff
@@ -0,0 +1,70 @@
...
+				{ethers.utils.formatEther(bid.yourBid.toString())} WILD
+			</td>
+			<td data-testid={TEST_ID.HIGHEST_BID}>
+				{ethers.utils.formatEther(bid.highestBid)} WILD
+			</td>
+			<td
+				data-testid={TEST_ID.STATUS}
+				className={isHighestBid ? styles.Lead : styles.Outbid}
+			>
+				{isHighestBid ? 'You Lead' : 'Outbid'}
```
ericlowturing, line 57: nit: It will be better to have this text as a constant and use it in both of this component and above test files.

```
const TEXTS = {
   STATUS: {
     NORMAL: 'Outbid',
    HIGHLIGHT: 'You Lead'
   }
};
```

So we can change the text in once place( here) instead of changing in both of component and test. (when we need to change the text)

src/containers/Tables/BidTable/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+						yourBid: ethers.utils.parseEther(bid.amount.toString()),
+						highestBid: highestBids.filter((d) => d.id === bid.tokenId)[0]
+							.amount,
+					};
+				})
+				.sort((a, b) => b.date.getTime() - a.date.getTime());
+		} catch (e) {
+			console.error(e);
+			throw new Error('Failed to parse bid data.');
+		}
```
ericlowturing, line 94: I think we can create a logical utility function here since it is just getting bidData from API.

```
const getBidsTableData = async (sdk, account) => {

.....

return bidData;
};
```

```
const getData = async () => {
  if (!account) {
    return;
  }

  setIsLoading(true);

 const tableData = await getBidsTableData(sdk, account);
};
```

And then we can create a test for this utility function.

src/containers/flows/CancelBid/CancelBid.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+	const cancelBidButton = await screen.findByRole('button', {
+		name: /cancel bid/i,
+	});
+	fireEvent.mouseUp(cancelBidButton);
+
+	await screen.findByText(constants.TEXT_CANCELLING_BID);
+
+	await screen.findByText('Error message');
+});
```
ericlowturing, line 207: Nice tests!

src/containers/flows/CancelBid/CancelBid.tsx
```diff
@@ -0,0 +1,125 @@
...
+	steps[Step.Cancelling] = (
+		<Wizard.Loading message={constants.TEXT_CANCELLING_BID} />
+	);
+	steps[Step.Success] = (
+		<Wizard.Confirmation
+			message={constants.TEXT_SUCCESS}
+			primaryButtonText={'Finish'}
+			onClickPrimaryButton={onFinish}
+		/>
+	);
```
ericlowturing, line 120: I have a question - I think it could be object instead of array?

```
const steps = {
   [Step.LoadingData]: <Wizard.Loading message={constants.TEXT_LOADING} />,
   ....
};
```

Just one time definition.

src/containers/Tables/BidTable/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+			setIsLoading(false);
+		}
+	};
+
+	const refetch = () => {
+		setBidData(undefined);
+		getData().catch((e) => {
+			if (e.message) {
+				// @todo handle errors in here
+				console.log(e.message);
```
ericlowturing, line 107: Agreed! I can see many of `console.log(error)`. We have `useNotification` which is similar `Toast`. I think we can style `notification` with `success, error, warning` and show `error` style message if we need to show message to user. Maybe we can ask notification styles to Eth :)

src/containers/Tables/BidTable/BidTableRow.test.tsx
colbr: I tried that, but the test suite wasn't happy rendering a `<tr>` outside of a `<table>` and `<tbody>` - bit annoying

src/containers/Tables/BidTable/BidTableRow.test.tsx
colbr: Good call, `toHaveTextContent` is probably easier to read that `toBe` in this situation

src/containers/Tables/BidTable/BidTableRow.test.tsx
colbr: Thanks!

src/containers/Tables/BidTable/BidTableRow.tsx
colbr: Good call. I also changed the `"WILD"` string to a constant in case we need to show a different token name at some point

src/containers/flows/CancelBid/CancelBid.test.tsx
colbr: Thanks!

src/containers/flows/CancelBid/CancelBid.tsx
colbr: True! Done

src/containers/Tables/BidTable/index.tsx
colbr: I think for now I'll just show the error message in the Cancel Bid model, but in the future we should for sure have a toaster component for long-running processes

PR #407, opened 2022-03-09
Switch StakingRequestProvider to Redux

ericlowturing:
[Associated Notion Card](https://www.notion.so/dcf93eee667d4edeac4b2a927d211551?v=0fbbeae3210b49af8266e3495e7d8ba5&p=4b326190d408403cbcb2dacad096fe99)

-----

## 1. Pull request checklist


src/lib/hooks/useStaking.tsx
```diff
@@ -0,0 +1,210 @@
+import { useMemo, useCallback } from 'react';
```
domw30, line 1: Leave this for now as we can do this in it's own PR - We should extract the error text to a global error constant file and replace with all instances in the codebase, I think we re-use them quite a bit.

domw30, line 1: Leave this for now as we can do this in it's own PR - We should extract the error text in the file to a global error constant file and replace with all instances in the codebase, I think we re-use them quite a bit.

src/lib/hooks/useStaking.tsx
```diff
@@ -0,0 +1,210 @@
...
+import {
+	createDomainMetadata,
+	UploadedDomainMetadata,
+	uploadToIPFS,
+} from 'lib/utils';
+import useNotification from 'lib/hooks/useNotification';
+import { useStakingController } from 'lib/hooks/useStakingController';
+import { useStakingRedux } from 'store/staking/hooks';
+
+export type UseStakingReturn = {
```
domw30, line 19: just a curious question mate...what is the difference using ```type``` over ```interface``` for the function props? I ask as I've been instructed in previous projects to use ```interface```

src/store/staking/actions.test.ts
```diff
@@ -0,0 +1,104 @@
...
+	setStakingFulfillingRequest,
+	setStakingFulfilledRequest,
+} from './actions';
+import {
+	NFT_STATUS_CARD_1,
+	NFT_STATUS_CARD_2,
+	DOMAIN_REQUEST_AND_CONTENTS_1,
+	DOMAIN_REQUEST_AND_CONTENTS_2,
+} from './staking.mockData';
+
```
domw30, line 31: Nice testing 🔥

src/store/staking/types.ts
```diff
@@ -0,0 +1,76 @@
...
+export type StakingState = {
+	requesting: NftStatusCard[];
+	requested: NftStatusCard[];
+	approving: DomainRequestAndContents[];
+	approved: DomainRequestAndContents[];
+	fulfilling: DomainRequestAndContents[];
+	fulfilled: DomainRequestAndContents[];
+};
+
+/**
```
domw30, line 23: I could be wrong but the comment and type names below look like they have slight typos? is ```stacking``` supposed to be ```staking```?

domw30 (approved):
Nice 👍  Approving with a couple of questions 😁

src/lib/hooks/useStaking.tsx
ericlowturing: @domw30 You are right. I sometimes forget it :(

src/store/staking/types.ts
ericlowturing: Good call - This is typo

PR #410, opened 2022-03-10
feat(accept-bid): handle pending and completed tx

colbr:


colbr:
Tested locally by @stacy-hunt

PR #411, opened 2022-03-11
Firefox browser backgrounds fix

domw30:
[Associated Notion Card](https://www.notion.so/dApp-Core-Product-Board-e96437225e264ae8ae8d46ca5f4d7b35?p=6a9d95a5d71b4e8fa615db1a4f25d810)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Bugfix

Elements with the ```blur``` class had a transparency issue on firefox browser.
Screen-Shot-2022-03-04-at-5-27-02-PM

Added ```background-primary``` to each element with the blur class

- [x] Tested on firefox

### DIFF
--- a/src/components/Banners/NextDrop/NextDrop.js
+++ b/src/components/Banners/NextDrop/NextDrop.js
@@ -30,7 +30,7 @@ const NextDrop = (props) => {
 	return (
 		<div
 			style={{ ...props.style }}
-			className={`${styles.nextDrop} border-primary border-rounded background-primary blur`}
+			className={`${styles.nextDrop} border-primary border-rounded background-primary`}
 		>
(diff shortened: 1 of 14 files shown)


colbr (changes requested):
Noticed a couple spots on NFTView are still transparent. Should be able to remove the `blur` class from all elements. It has an `!important` flag so it will override `background-primary`.

Screen Shot 2022-03-14 at 11 05 11 AM

domw30:
@colbr, my bad thanks for that. The ```blur``` class didn't seem to override the style when I was testing on BrowserStack, but then again I was limited to a minute on each ```device``` so quite rushed.  I've removed any use of blur class from all elements 👍

colbr (approved):
Nice one, all looking good to me

PR #412, opened 2022-03-11
Investigate and resolve ‘Cache data may be lost when replacing domains field of a Query object’ error

domw30:
[Associated Notion Card](https://www.notion.so/dApp-Core-Product-Board-e96437225e264ae8ae8d46ca5f4d7b35?p=3d3ca539b413421d9760edf557cce130)

-----

## 1. Pull request checklist

## 2. PR type
- [x] bug/warning fix

Cache data may be loss - existing and incoming 

Screenshot 2022-03-11 at 11 32 59

Screenshot 2022-03-10 at 21 33 34

Warning resolved by adding typePolicies to InMemoryCache

My Findings: 
From my understanding, this warning occurs because there is no explicit id or _id property and in Apollo Client 3 the inMemoryCache does not create a fallback fake identifier to normalise the data.

From [Apollo Docs](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#configuring-merge-functions-for-types-rather-than-fields): 

*If you define a merge function for a field, the cache calls that function whenever the field is about to be written with an incoming value (such as from your GraphQL server). When the write occurs, the field's new value is set to the merge  function's return value, instead of the original incoming value.*

*Your merge function cannot push the incoming array directly onto the existing array. It must instead return a new array to prevent potential errors.*

### DIFF
--- a/src/lib/providers/SubgraphProvider.tsx
+++ b/src/lib/providers/SubgraphProvider.tsx
@@ -42,7 +42,17 @@ export const SubgraphProvider: React.FC<SubgraphContextProviderType> = ({
 	const apolloClient = React.useMemo(() => {
 		return new ApolloClient({
 			uri: chainIdToSubgraph(chainSelector.selectedChain),
-			cache: new InMemoryCache(),
+			cache: new InMemoryCache({
+				typePolicies: {
+					Query: {
+						fields: {
+							domains: {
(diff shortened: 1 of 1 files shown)


colbr (approved):
Makes sense! Thanks for picking this up.

PR #419, opened 2022-03-16
Fix to show bid button by setting isBiddable as true as default

ericlowturing:
[Associated Notion Card](https://www.notion.so/Fix-missing-row-button-c0cdb0a1b6714057b9becd4c0c4fcc86)

-----

## 1. Pull request checklist


colbr (approved):
Nice, makes sense. FYI the `CancelBid` modal uses the `getMetadata` function from the SDK rather than the one defined in our app, but that modal doesn't react to any of the advanced metadata so that will be ok.

PR #428, opened 2022-03-23
feat: Allow retrieval of buy now price when no wallet connected

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Fix-no-buy-now-price-when-wallet-disconnected-UX-12b5b9c0806e498998d612232f10a1f8)

-----

## 1. Pull request checklist

## 2. PR type
 

Bugfix/feature

Buy Now price could not be retrieved through SDK without a wallet connected for two reasons:
1. The SDK `getBuyNowPrice` method required a `provider` as a parameter. When the user isn't connected, there is no `provider` to use. Instead, the SDK should use the `provider` passed to it during configuration in `ZnsSdkProvider`
2. No default `provider` was being created when creating ZNS SDK instance, so we were sending it `undefined`

1. Updated ZNS SDK - now `getBuyNowPrice` uses the `provider` given during configuration.
2. Created a default `provider` for ZNS SDK configuration if no wallet is connected.

## 5. Other information

Also did some minor refactoring in `ZnsSdkProvider` as follows:
1. Used `NETWORK_TYPES` instead of `chainId` when creating SDK instance - this makes it a little bit easier to read.
2. Removed all the unused functions (i.e. all of them)


padala (approved):

padala:
@colbr looks good to me

PR #429, opened 2022-03-23
feat: static "Create A DAO" flow

joonchoi0918:
[Associated Notion Card](https://www.notion.so/zerotech/Make-static-Create-DAO-flow-a12ca1c2210149e2a0a763e1b12b346f)

-----

## 1. Pull request checklist

## 2. PR type
 

Feature

There was no way to create a DAO.

Added a modal for creating a new DAO per Figma.

## 5. Other information

This is currently just a static element - real data will be used when SDK is integrated.

### DIFF
--- a/src/containers/DAO/pages/MintDAO/MintDAO.tsx
+++ b/src/containers/DAO/pages/MintDAO/MintDAO.tsx
@@ -1,16 +1,26 @@
 import { useMemo, useState } from 'react';
-import { Overlay, FutureButton, Wizard } from 'components';
+import { Overlay, Wizard } from 'components';
 import { useManyClickHandlers } from 'lib/hooks/useManyClickHandlers';
 import { useUpdateEffect } from 'lib/hooks/useUpdateEffect';
 import useNotification from 'lib/hooks/useNotification';
-import { Step, StepStatus, StepWizard, MintData } from './MintDAO.types';
+import {
+	Step,
+	StepStatus,
+	StepWizard,
+	MintData,
+	ErrorType,
+} from './MintDAO.types';
 import { ERRORS, SUCCESS } from './MintDAO.constants';
 import { formatMintWizard } from './MintDAO.helpers';
 
-export const MintDAO: React.FC = () => {
+interface MintDAOProps {
+	onClose: () => void;
+}
+
+export const MintDAO: React.FC<MintDAOProps> = ({ onClose }) => {
 	const { addNotification } = useNotification();
 
-	const [currentStep, setCurrentStep] = useState<Step>(Step.None);
+	const [currentStep, setCurrentStep] = useState<Step>(Step.Unlock);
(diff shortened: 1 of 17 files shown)


src/containers/DAO/pages/MintDAO/MintDAO.tsx
```diff
@@ -0,0 +1,127 @@
...
+	);
+	const [currentMintData, setCurrentMintData] = useState<
+		MintData | undefined
+	>();
+	const [error, setError] = useState<string | undefined>();
+
+	const onChangeStep = (step: Step) => () => setCurrentStep(step);
+
+	const onChangeStepStatus = (stepStatus: StepStatus) => () =>
+		setCurrentStepStatus(stepStatus);
```
colbr, line 25: I like the currying 😊

src/containers/DAO/pages/MintDAO/MintDAO.types.ts
```diff
@@ -0,0 +1,27 @@
...
+	Mint = 'Mint',
+}
+
+export enum StepStatus {
+	Normal = 'Normal',
+	Confirm = 'Confirm',
+	Processing = 'Propcessing',
+}
+
+export enum ErrrorType {
```
colbr, line 13: Small type - should be `ErrorType`

src/containers/DAO/pages/MintDAO/MintDAO.types.ts
```diff
@@ -0,0 +1,27 @@
...
+	Signature = 'Signature',
+	Transaction = 'Transaction',
```
colbr, line 15: This flow won't be generating any signatures, so can remove the `Transaction` error type

src/containers/DAO/pages/MintDAO/MintDAO.tsx
```diff
@@ -0,0 +1,127 @@
...
+			{wizard && (
+				<Overlay
+					open
+					onClose={onClose}
+					onContentClick={onMockStepResponse}
+					hasCloseButton={currentStepStatus === StepStatus.Normal}
+				>
+					<Wizard header={wizard.title}>{wizard.children}</Wizard>
+				</Overlay>
+			)}
```
colbr, line 122: Can we separate this logic out so that `MintDAO.tsx` is just the modal itself, and then make a `MintDAOButton.tsx` component which has a button which toggles the wizard/modal on and off? Similar to have `CancelBid.tsx` and `CancelBidButton.tsx` work.

This is because we have to trigger the `<MintDAO>` modal from a few different places in the app, so will need to be able to separate the modal from the button triggering it.

colbr (changes requested):
Looking awesome! Couple of small requests in here.

Could I also get you to write some UI tests for the flow? Similar to `CancelBid.test.tsx`.

Nice work @joonchoi0918 - killing it

colbr (approved):

PR #447, opened 2022-03-29
Fix/biddable issue

eric-zero-dev:
[Associated Notion Card](https://www.notion.so/zerotech/Shouldn-t-be-able-to-bid-on-NFTs-with-bidding-disabled-730698b2d8e7452fb8c42adc6251e8c8)

-----

## 1. Pull request checklist

### DIFF
--- a/src/containers/flows/MakeABid/MakeABid.tsx
+++ b/src/containers/flows/MakeABid/MakeABid.tsx
@@ -429,7 +429,7 @@ const MakeABid: React.FC<MakeABidProps> = ({ domain, onBid }) => {
 					)}
 					{!isDomainOwner && !isBiddable && (
 						<p className={styles.Error} style={{ paddingTop: '16px' }}>
-							You can not bid on this domain because it is disabled bid
+							Bidding has been disabled on this domain
 						</p>
 					)}
 					{!isDomainOwner && isBiddable && (
--- a/src/containers/flows/MakeABid/MakeABid.tsx
+++ b/src/containers/flows/MakeABid/MakeABid.tsx
@@ -83,6 +83,17 @@ const MakeABid: React.FC<MakeABidProps> = ({ domain, onBid }) => {
 	const zAuctionAddress = znsContracts.zAuction.address;
 	const wildContract: ERC20 = znsContracts.wildToken;
 
+	const { isBiddable, isDomainOwner } = useMemo(() => {
+		const isBiddable = Boolean(domainMetadata?.isBiddable);
+		const isDomainOwner =
+			domain.owner.id.toLowerCase() === account?.toLowerCase();
+
+		return {
+			isBiddable,
+			isDomainOwner,
+		};
+	}, [domain, domainMetadata, account]);
+
 	const isBidValid = useMemo(() => {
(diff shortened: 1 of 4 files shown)


src/containers/other/NFTView/elements/NFT/NFT.tsx
```diff
@@ -92,6 +92,7 @@ const NFT = ({
 	onRefetch,
 }: NFTProps) => {
 	const blobCache = useRef<string>();
+	console.log({ isBiddable });
```
domw30, line 95: is this to be removed?

src/containers/flows/MakeABid/MakeABid.tsx
```diff
@@ -407,12 +422,17 @@ const MakeABid: React.FC<MakeABidProps> = ({ domain, onBid }) => {
...
-					{domain.owner.id.toLowerCase() === account?.toLowerCase() && (
+					{isDomainOwner && (
 						<p className={styles.Error} style={{ paddingTop: '16px' }}>
 							You can not bid on your own domain
 						</p>
 					)}
-					{domain.owner.id.toLowerCase() !== account?.toLowerCase() && (
+					{!isDomainOwner && !isBiddable && (
+						<p className={styles.Error} style={{ paddingTop: '16px' }}>
+							You can not bid on this domain because it is disabled bid
```
domw30, line 432: I think we should change this slightly if you agree...

You can not bid on this domain because it is disabled bid -> Bidding has been disabled on this domain

domw30 (approved):
Eric looks good thanks! 

Domain Bidding still looks like it is set as off by default, this causes the Make A Bid buttons to disappear on load

src/containers/other/NFTView/elements/NFT/NFT.tsx
eric-zero-dev: Yeah, just removed!

src/containers/flows/MakeABid/MakeABid.tsx
```diff
@@ -83,6 +83,17 @@ const MakeABid: React.FC<MakeABidProps> = ({ domain, onBid }) => {
 	const zAuctionAddress = znsContracts.zAuction.address;
 	const wildContract: ERC20 = znsContracts.wildToken;
 
+	const { isBiddable, isDomainOwner } = useMemo(() => {
+		const isBiddable = Boolean(domainMetadata?.isBiddable);
```
colbr, line 87: Change this logic to `domainMetadata?.isBiddable ?? true`, as `undefined` `isBiddable` should default to `true`

src/containers/flows/MakeABid/MakeABid.tsx
eric-zero-dev: Updated the text

colbr (changes requested):

src/containers/flows/MakeABid/MakeABid.tsx
colbr: Update: talking about this in a different thread - Eric is going to change the metadata retrieval logic to handle this

PR #452, opened 2022-03-30
feat(domain-search): fix issue with domain search

padala:
[0.13.0 Release testing](https://www.notion.so/zerotech/0-13-0-Release-testing-88348931b0234f008deb032382fdfbb4)

-----

## 1. Pull request checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,6 +134,7 @@
 		"husky": "^4.3.8",
 		"lint-staged": "^11.1.2",
 		"prettier": "^2.2.1",
+		"react-error-overlay": "^6.0.9",
 		"ts-essentials": "^7.0.1",
 		"typechain": "^4.0.3"
 	},
--- a/src/lib/useDomainSearch.tsx
+++ b/src/lib/useDomainSearch.tsx
@@ -20,31 +20,39 @@ export function useDomainSearch() {
 
 	useAsyncEffect(async () => {
 		const id = getDomainId(pattern);
-		try {
-			const rawDomain = await sdk.getDomainById(id);
-			const exactDomain = rawDomain as any;
-			exactDomain.metadata = exactDomain.metadataUri;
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -134,6 +134,7 @@
 		"husky": "^4.3.8",
 		"lint-staged": "^11.1.2",
 		"prettier": "^2.2.1",
+		"react-error-overlay": "^6.0.9",
```
padala, line 137: An issue with hot reload. More details here - https://github.com/facebook/create-react-app/issues/11773

src/lib/useDomainSearch.tsx
```diff
@@ -20,31 +20,39 @@ export function useDomainSearch() {
...
-			const rawDomain = await sdk.getDomainById(id);
-			const exactDomain = rawDomain as any;
-			exactDomain.metadata = exactDomain.metadataUri;
-			exactDomain.minter = { id: exactDomain.minter };
-			exactDomain.owner = { id: exactDomain.owner };
-			exactDomain.parent = { id: exactDomain.parentId };
-			exactDomain.lockedBy = { id: exactDomain.lockedBy };
+		if (id) {
+			try {
+				const rawDomain = await sdk.getDomainById(id);
```
colbr, line 25: Maybe not necessary for this iteration, but I think we could speed this up by removing `sdk.getDomainById` for exact match.

Something like:
```
get domains by name
filter domains for exact match of search query
```

This way there would be one API call per search query rather than two.

colbr (approved):
Looks good. Added some comments re: one optimisation, but not necessary given the timeframe before release.

src/lib/useDomainSearch.tsx
padala: @colbr  Yeah makes sense. We can do that. I also think we need to put this behind a debounce to avoid frequent calls as user types ahead.

PR #453, opened 2022-03-30
fix(buy-now): Use standardised getMetadata method

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Metadata-not-being-parsed-in-two-spots-b6e0df1465f34ac281d9d72e517e7829)

-----

## 1. Pull request checklist

## 2. PR type
 

Bugfix/refactoring

`SetBuyNow` and `BuyNow` modals weren't properly parsing metadata.

`SetBuyNow` and `BuyNow` now use `getMetadata` to retrieve metadata.

### DIFF
--- a/src/containers/flows/BuyNow/index.tsx
+++ b/src/containers/flows/BuyNow/index.tsx
@@ -15,7 +15,7 @@ import { Data } from './BuyNow';
 import { useZnsContracts } from 'lib/contracts';
 import { ERC20 } from 'types';
 import { ethers } from 'ethers';
-import { getMetadata } from 'lib/metadata';
+import useMetadata from 'lib/hooks/useMetadata';
 
 export type BuyNowContainerProps = {
 	domainId: string;
@@ -30,6 +30,7 @@ const BuyNowContainer = ({
 }: BuyNowContainerProps) => {
 	// Hooks
 	const { instance: sdk } = useZnsSdk();
+	const { getMetadata } = useMetadata();
 	const { account, library } = useWeb3React();
 	const { wildPriceUsd } = useCurrency();
 	const { addNotification } = useNotification();
--- a/src/containers/flows/BuyNow/index.tsx
+++ b/src/containers/flows/BuyNow/index.tsx
@@ -15,6 +15,7 @@ import { Data } from './BuyNow';
 import { useZnsContracts } from 'lib/contracts';
 import { ERC20 } from 'types';
 import { ethers } from 'ethers';
+import { getMetadata } from 'lib/metadata';
 
 export type BuyNowContainerProps = {
 	domainId: string;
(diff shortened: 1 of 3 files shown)


src/containers/flows/BuyNow/index.tsx
```diff
@@ -143,11 +144,11 @@ const BuyNowContainer = ({
...
 		}
 		try {
-			const [domain, metadata, balance] = await Promise.all([
+			const [domain, balance] = await Promise.all([
 				sdk.getDomainById(domainId),
-				sdk.getDomainMetadata(domainId, library.getSigner()),
 				wildContract.balanceOf(account),
 			]);
+			const metadata = await getMetadata(domain.metadataUri);
 			if (domain && metadata && buyNowPrice && balance) {
```
padala, line 152: @colbr  Can we use `sdk.utility.getMetadataFromUri(uri)` instead of using `getMetadata(uri)`. As we wanted to get rid of all the direct interactions.

colbr: Unfortunately metadata has to be parsed to process default values properly. For example, the SDK will return `isBiddable: undefined`, which means wherever we use it we have to do `isBiddable ?? true` since `isBiddable` should default to `true` if it's `undefined`.

I'll make some cleaner code for this, let me do another push

padala: Yeah makes sense. Probably a transform method. Sure, I'll wait for your changes.

domw30 (approved):
nice 👍

src/lib/hooks/useMetadata.ts
```diff
@@ -0,0 +1,36 @@
...
+ * are not being set in the SDK - for example, domains should be biddable
+ * by default, so isBiddable = undefined should be isBiddable = true
+ *
+ * @returns function for getting metadata through SDK
+ */
+const useMetadata = (): UseMetadataReturn => {
+	const { instance: sdk } = useZnsSdk();
+
+	const getMetadata = async (uri: string) => {
+		const raw = await sdk.utility.getMetadataFromUri(uri);
```
eric-zero-dev, line 25: nit: Let's import `import { DomainMetadata } from '@zero-tech/zns-sdk/lib/types';` and use `const raw: DomainMetadata = await sdk.utility.getMetadataFromUri(uri);`, so that we can know it is converting SDK's `DomainMetadata` to dApp's `Metadata `;

src/lib/hooks/useMetadata.ts
```diff
@@ -0,0 +1,36 @@
...
+		if (raw) {
+			return parseDomainMetadata(raw);
+		}
+	};
+
+	return {
+		getMetadata,
+	};
```
eric-zero-dev, line 33: Looks great! I think we can add `setMetadata` and use it in `Edit Metadata` flow. So `getter/setter` with `hooks`. Maybe we can do it later👍

eric-zero-dev (approved):
Great work!

PR #454, opened 2022-03-30
refactor: Use context API for SDK use

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Use-Context-API-for-SDK-9eab3ba31e114d009d5d528567a20e70)

-----

## 1. Pull request checklist

## 2. PR type
 

Refactoring

The SDK was just inside a custom hook in a `useMemo`, with the intention of the `useMemo` memo-ising the SDK instance and returning the same object every time it's called. However, creating a new instance of `useZnsSdk` also creates a new instance of the `useMemo`, and so the SDK gets reinitialised.

We're using context API for the SDK now so it just gets initialised once, rather than every time it's needed.

### DIFF
--- a/src/lib/providers/ZnsSdkProvider.tsx
+++ b/src/lib/providers/ZnsSdkProvider.tsx
@@ -1,0 +1,70 @@
+import { useWeb3React } from '@web3-react/core';
+import * as zns from '@zero-tech/zns-sdk';
+import React from 'react';
+import { useChainSelector } from 'lib/providers/ChainSelectorProvider';
+import { ethers } from 'ethers';
+import {
+	chainIdToNetworkType,
+	defaultNetworkId,
+	NETWORK_TYPES,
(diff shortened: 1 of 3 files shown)


src/lib/providers/ZnsSdkProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type ZnsSdkProviderProps = {
+	children: React.ReactNode;
+};
+
+export const SdkContext = React.createContext({
+	instance: {} as zns.Instance,
+});
+
+export const ZnsSdkProvider = ({ children }: ZnsSdkProviderProps) => {
```
eric-zero-dev, line 21: nit: `ZnsSdkProvider: React.FC`

eric-zero-dev (approved):
Looks Good To Me. Nice working!

PR #474, opened 2022-04-04
feat: Improvements to NFT View

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=c2d96f08ab6c4773981b9d9eef82c740)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Feature

No view bids button on NFT view page.
Only two dropdown options.

View bids button on NFT view page and conditions when to show.
Four dropdown options as per design.

**LOOM WALKTHROUGH**
[LOOM WALKTHROUGH](https://www.loom.com/share/e51cf08008b44336bca970f364d0442a)


src/containers/other/NFTView/NFTView.tsx
```diff
@@ -142,6 +157,14 @@ const NFTView: React.FC<NFTViewProps> = ({ onTransfer }) => {
 				wildPriceUsd={wildPriceUsd}
 				refetch={refetch}
 				isBiddable={isBiddable}
+				bidData={allBids}
+				isLoading={isBidDataLoading}
+				domainMetadata={domainMetadata}
+				domain={domainData}
+				setIsViewBidsOpen={setIsViewBidsOpen}
```
domw30, line 164: Maybe it would be better to pull the Overlays up to NFTView and handle it from there rather than passing more props down?

src/containers/other/NFTView/NFTView.tsx
```diff
@@ -22,6 +22,7 @@ import {
 import useCurrency from 'lib/hooks/useCurrency';
```
domw30, line 22: NFTView appears to have a data leak but I've been unable to identify exactly where that is happening!

domw30:
Hey @colbr, please could I get your feedback on this PR. I've left a couple of comments - triggering the modals the way its currently working feels a bit hacky, what d'ya think?

src/containers/buttons/ViewBidsButton/ViewBidsButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+	return (
+		<>
+			{isViewBidsOpen && domainMetadata && bids && (
+				<Overlay open={isViewBidsOpen} centered onClose={onClose}>
+					<BidList
+						bids={bids}
+						domain={domain}
+						domainMetadata={domainMetadata}
+						onAccept={onAccept}
+						highestBid={String(highestBid)}
```
domw30, line 70: Struggling to format the ```highestBid``` value. When using existing format utils it will crash the browser.
Screenshot 2022-04-04 at 21 39 16

src/containers/buttons/ViewBidsButton/ViewBidsButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Maybe, Metadata } from 'lib/types';
+import { Domain } from '@zero-tech/zns-sdk/lib/types';
+
+//- Components Imports
+import { FutureButton, Overlay, TextButton } from 'components';
+
+//- Constants Imports
+import { LABELS } from './ViewBidsButton.constants';
+
+interface ViewBidButtonProps {
```
colbr, line 15: Just a thought - we could also make a `ViewBidButton` which just takes a `domainId` as props and gets all the modal data internally. Would cost an extra API call, but could reduce complexity of code as we won't need to pass in a tonne of data. Not necessary for this task though as this seems to be working sweet.

src/containers/buttons/ViewBidsButton/ViewBidsButton.tsx
colbr: It looks like the `<Details>` component in `AcceptBid` is looking for bid amount in wei as a string. Should be able to send it as `parseEther(highestBid.toString()).toString()`. Could be worth changing the variable to `highestBidAsWei`.

src/containers/flows/SetBuyNow/SetBuyNowButton.tsx
```diff
@@ -18,23 +19,26 @@ const SetBuyNowButton = ({
...
+			{isSetBuyNowOpen && domainId && (
+				<Overlay
+					open={isSetBuyNowOpen}
+					onClose={() => setIsSetBuyNowOpen && setIsSetBuyNowOpen(false)}
+				>
 					<SetBuyNow
 						onSuccess={onSuccess}
 						domainId={domainId}
-						onCancel={() => setIsModalOpen(false)}
+						onCancel={() => setIsSetBuyNowOpen && setIsSetBuyNowOpen(false)}
```
colbr, line 41: nit: could be  `onCancel={() => setIsSetBuyNowOpen?.(false)}`

src/containers/other/NFTView/NFTView.tsx
colbr: Looks like `useNftData` isn't checking if it's still mounted before setting state. Will just have to add an `isMounted` ref to the hook, and set it on mount/unset it on dismount. Have a look at `useBidData` in `CancelBid` for an example.

src/containers/other/NFTView/NFTView.tsx
colbr: It could even be worth making a hook/provider with Context API for controlling the modals - would reduce complexity at the cost of slightly tighter coupling. Just one called `NFTViewModalProvider` to use locally on the `NFTView` component would be quite handy, as there's a *lot* of prop passing going on here.

However, I don't think you'll need to pass down state for Buy Now and Set Buy Now, as the wrapper buttons should come with internal show/hide modal logic.

src/containers/buttons/ViewBidsButton/ViewBidsButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+	buttonText,
+	className,
+	isTextButton,
+	refetch,
+	style,
+	highestBid,
+	isLoading,
+	domainMetadata,
+	setIsViewBidsOpen,
+	isViewBidsOpen,
```
colbr, line 42: Generally when I make overlay buttons I would define these internally, as the purpose of an overlay button is to encapsulate all the functionality of showing/hiding the modal. I don't think you'd need to control the modal rendering from outside of this component, but I could be wrong.

src/containers/other/NFTView/elements/Actions/Actions.tsx
```diff
@@ -68,83 +85,103 @@ const Actions = ({
...
 			buttonComponent: (isTextButton?: boolean) => (
 				<SetBuyNowButton
 					onSuccess={refetch}
-					buttonText={buyNowPrice ? 'Edit Buy Now' : 'Set Buy Now'}
+					buttonText={buyNowPrice ? LABELS.EDIT_BUY_NOW : LABELS.SET_BUY_NOW}
 					domainId={domainId ?? ''}
 					isTextButton={isTextButton}
 					className={cx({ TextButton: isTextButton })}
+					setIsSetBuyNowOpen={setIsSetBuyNowOpen}
+					isSetBuyNowOpen={isSetBuyNowOpen}
```
colbr, line 158: Said above, but shouldn't need to pass these values in as modal rendering is controlled within the `SetBuyNowButton` component (`ViewBidsButton` should behave this way too)

src/containers/other/NFTView/elements/Actions/Actions.tsx
```diff
@@ -165,7 +223,7 @@ const Actions = ({
 	 */
 	const ordered = isOwnedByUser
 		? [
-				actions[ACTION_TYPES.PlaceBid],
```
colbr, line 168: Could reduce complexity by just having a `Bid` action type, and the `buttonComponent` in its `ActionBlock` changes based on ownership. Would require one less `ActionBlock` object above.

src/containers/other/NFTView/hooks/useNftData.ts
```diff
@@ -12,6 +12,7 @@ import { useCurrentDomain } from 'lib/providers/CurrentDomainProvider';
 import useNotification from 'lib/hooks/useNotification';
```
colbr, line 12: This hook is getting pretty heavy - definitely need to split it out into grouped functionality at some point, but maybe outside the scope of this task!

colbr (changes requested):
Looks really nice, good job.

Few changes, mainly:
- Adjust buttons which open modals as per comments
- Pass wei amount as `highestBid` for correct formatting

src/containers/other/NFTView/NFTView.tsx
```diff
@@ -121,27 +188,29 @@ const NFTView: React.FC<NFTViewProps> = ({ onTransfer }) => {
...
-				onSelectOption={onSelectNFTMoreOption}
+				onSelectOption={onSelectOption}
 			/>
 
 			<Actions
 				domainId={znsDomain?.id}
 				highestBid={highestBid}
 				buyNowPrice={buyNowPrice}
+				onMakeBid={openMakeABid}
+				onViewBids={openBidList}
```
domw30, line 207: It would be nice to do this for each of the buttons but if the buttons are re-used in the codebase they would require a modal provider.

domw30:
@colbr, made some major changes to this based on your feedback. 

All the feedback tasks I have worked on have been listed in iteration 1 in notion:

https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=c2d96f08ab6c4773981b9d9eef82c740

src/containers/other/NFTView/elements/Stats/Stats.tsx
```diff
@@ -43,6 +49,7 @@ export const Stats: React.FC<StatsProps> = ({
 	bids,
 	isLoading,
 }) => {
+	const isMounted = useRef<boolean>();
```
domw30, line 52: I believe this was the culprit for the react state update on unmounted component console warning

src/containers/other/NFTView/providers/NFTViewModalProvider/NFTViewModalProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+						domain={modalContent.contentProps.domain}
+						domainMetadata={modalContent.contentProps.domainMetadata}
+						onAccept={modalContent.contentProps.onAccept}
+						highestBid={modalContent.contentProps.highestBid}
+						isLoading={modalContent.contentProps.isLoading}
+					/>
+				)}
+			</Overlay>
+
+			{modalContent?.modalType === NFTViewModalType.DOMAIN_SETTINGS && (
```
domw30, line 74: Overlay is not wrapped around DomainSettings here as it already has an Overlay in the component

src/containers/other/NFTView/providers/NFTViewModalProvider/NFTViewModalProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+				)}
+				{modalContent?.modalType === NFTViewModalType.BID_LIST && (
+					<BidList
+						bids={modalContent.contentProps.bids}
+						domain={modalContent.contentProps.domain}
+						domainMetadata={modalContent.contentProps.domainMetadata}
+						onAccept={modalContent.contentProps.onAccept}
+						highestBid={modalContent.contentProps.highestBid}
+						isLoading={modalContent.contentProps.isLoading}
+					/>
```
domw30, line 70: Reminder to add ```isAcceptBidEnabled``` prop when rebasing with develop - filters bids.

colbr (approved):
Looking pretty good to me - since it's a big review I may get Eric, Subbu, or Joon to have a look as wel!

src/containers/other/NFTView/NFTView.tsx
```diff
@@ -77,40 +83,107 @@ const NFTView: React.FC<NFTViewProps> = ({ onTransfer }) => {
...
+				return onTransfer();
+			case NFT_MORE_ACTIONS_TITLE.SET_BUY_NOW:
+				return openSetBuyNow();
+			case NFT_MORE_ACTIONS_TITLE.EDIT_BUY_NOW:
+				return openSetBuyNow();
+			case NFT_MORE_ACTIONS_TITLE.VIEW_BIDS:
+				return openBidList();
+			default:
+				return '';
 		}
```
eric-zero-dev, line 187: I think `onSelectOption: (option: Option) => void;` so we don't need to return for `default` case.

src/containers/other/NFTView/hooks/useAsset.ts
```diff
@@ -0,0 +1,67 @@
...
+			return;
+		}
+
+		const asset = await getDomainAsset(domainAssetURL);
+
+		if (asset) {
+			addNotification(MESSAGES.DOWNLOAD);
+
+			await downloadDomainAsset(asset);
+		}
```
eric-zero-dev, line 51: nit: We can get failure on getDomainAsset & downloadDomainAsset. So it would be nice to have try/catch to error handling here.

src/containers/other/NFTView/providers/NFTViewModalProvider/NFTViewModalProvider.tsx
eric-zero-dev, line 70: nit: it is good time to use `switch` instead of `if/else`?

src/containers/other/NFTView/providers/NFTViewModalProvider/NFTViewModalProvider.types.ts
```diff
@@ -0,0 +1,74 @@
...
+	contentProps: BidListContentProps;
+}
+
+export type NFTViewModalContent =
+	| NullContent
+	| MakeABidContent
+	| DomainSettingsContent
+	| TransferOwnershipContent
+	| SetBuyNowContent
+	| BidListContent;
```
eric-zero-dev, line 68: Sweet!

eric-zero-dev (changes requested):
Looks good to me. I left a few nits to change. Nice work @domw30 I appreciate it. Thanks

domw30:
@eric-zero-dev, changes made, thanks! 😁

eric-zero-dev (approved):
Looks Good. Thanks @domw30

PR #483, opened 2022-04-06
hotfix: Filter Bids in Bid List when used can Accept Bids - Display all by default

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=fea194c799784b5488ee94c643d879ef)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Hotfix

- User is able to select bids they have made and open the accept bid flow.
- It is not obviously indicated that certain bids are user previous bids

- When a user can Accept Bids from the Bid List, we should filter out their own bids.
- When a user is not able to Accept Bids from the Bid List, we should not filter out their own bids.
- Text to indicate users bids

[Loom Walk through](https://www.loom.com/share/c0aef17a67f04b3db2a5815095070a1e)

Screenshot 2022-04-06 at 13 40 14

### DIFF
--- a/src/containers/lists/BidList/BidList.tsx
+++ b/src/containers/lists/BidList/BidList.tsx
@@ -59,10 +59,11 @@ const BidList: React.FC<BidListProps> = ({
 
 	// Sort bids by date
 	const sortedBids = sortBidsByTime(bids);
-	const filteredBids = sortedBids.filter(
-		(bid) => bid.bidder.toLowerCase() !== account?.toLowerCase(),
-	);
-	const isFilteredBids = isAcceptBidEnabled ? filteredBids : sortedBids;
+	const bidsToShow = isAcceptBidEnabled
+		? sortedBids.filter(
+				(bid) => bid.bidder.toLowerCase() !== account?.toLowerCase(),
+		  )
+		: sortedBids;
 
 	useEffect(() => {
 		if (library) {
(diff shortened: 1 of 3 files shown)


src/containers/lists/BidList/BidList.tsx
```diff
@@ -43,17 +47,22 @@ const BidList: React.FC<BidListProps> = ({
...
 	const [isAcceptBidModal, setIsAcceptBidModal] = useState(false);
 	const [acceptingBid, setAcceptingBid] = useState<Bid | undefined>(undefined);
-	const { library } = useWeb3React<Web3Provider>();
+	const { library, account } = useWeb3React<Web3Provider>();
 
 	// Sort bids by date
 	const sortedBids = sortBidsByTime(bids);
+	const filteredBids = sortedBids.filter(
+		(bid) => bid.bidder.toLowerCase() !== account?.toLowerCase(),
+	);
```
colbr, line 64: tiny nit: would be more efficient to only do the filtering if `isAcceptBid` is enabled - not a big deal as I don't imagine the lists will get that long, but something to consider!

src/containers/lists/BidList/BidList.tsx
```diff
@@ -43,17 +47,22 @@ const BidList: React.FC<BidListProps> = ({
...
+	const isFilteredBids = isAcceptBidEnabled ? filteredBids : sortedBids;
```
colbr, line 65: tiny nit: `isFilteredBids` implies the variable is a boolean (because of `is`) - I think it would be clearer to call it `filteredBid`, `bidsToShow`, or something similar.

colbr (approved):
Looks good! Nice work 🚀

Keep this branch alive just in case we need to merge it as a hotfix before the new UI drops

PR #499, opened 2022-04-12
fix: bid modal not loading, bid modal not restarting on wallet change

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Loading-Bid-Data-infinite-loading-loop-error-69cf72ef21ce46bab8101a583887e2f0)

-----

## 1. Pull request checklist

## 2. PR type
 

Bugfix

- Bid modal was getting stuck in an infinite loop if the connected wallet had 0 balance. 
- Bid modal would not start from the start if the connect wallet changed while the modal was open, meaning a user could accidentally place a bid without approving zAuction.
- The text while loading the "Place Bid" step in the modal is "Loading Bid Data", when it should be "Loading Domain Data"

- Bid modal now progresses with an error message if the user has 0 WILD. 
- The flow will also start from the start on wallet change.
- Modal loading text now says "Loading Domain Data"

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "zns-dapp",
-	"version": "0.14.2",
+	"version": "0.14.3",
 	"private": true,
 	"dependencies": {
 		"@apollo/client": "^3.3.13",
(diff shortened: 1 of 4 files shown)


colbr:
Approved in Notion

PR #508, opened 2022-04-13
hotfix: Cloudinary 404

colbr:
Removed Cloudinary version number from template URL, as it was giving some 404s


stacy-hunt (approved):
Yep, loading from Cloudinary now

PR #512, opened 2022-04-14
feat(mint-drop): handle wallet not connected scenario

padala:
Hotfix - fixes issue where user eligibility was inaccurate depending on when the user connected their wallet


colbr (approved):
Confirmed I can not replicate the same bug as before!

PR #542, opened 2022-05-01
fix: some assets not using Cloudinary

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/neo-In-general-seems-to-be-loading-very-slow-7ac7b2b966754790b875a4a3b23987df)

-----

## 1. Pull request checklist

## 2. PR type
 

Bugfix

1. Table cards weren't using Cloudinary, and instead were loading assets straight from IPFS. This was incredibly slow, for obvious reasons.
2. Table card images weren't being vertically aligned.

1. Table cards now use Cloudinary to load assets.
2. Table card images are now vertically aligned.

### DIFF
--- a/src/components/Cards/ImageCard/ImageCard.module.scss
+++ b/src/components/Cards/ImageCard/ImageCard.module.scss
@@ -55,7 +55,7 @@ $placeholder-background: #000;
 		position: absolute !important;
 		top: 0;
 		left: 0;
-		height: auto !important;
+		height: 100% !important;
 		width: 100% !important;
 		object-fit: cover;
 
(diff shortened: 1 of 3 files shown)


src/components/Cards/ImageCard/ImageCard.module.scss
```diff
@@ -55,7 +55,7 @@ $placeholder-background: #000;
 		position: absolute !important;
 		top: 0;
 		left: 0;
-		height: auto !important;
+		height: 100% !important;
```
domw30, line 58: Yeah nice this does make the image position centred on the preview page. Tricky handling the different image heights in grid view across all domains. Looks good 👍

domw30 (approved):
Yeah looks good, I did this too and noticed the difference in loading time but wasn't sure it would work for all uses ImageCard but seems all good my end. 👍

PR #543, opened 2022-05-01
bug: fix for metadata link in NFT View routing 

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=4bbbcd19ad9045c980f2873d5e094ad0)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Bug

Metadata links on nft view page were not redirecting to correct url

Redirect now goes to correct url

https://user-images.githubusercontent.com/39112648/166142824-fbdf296c-ea34-46d0-ad62-8fd4a3155ec2.mov

### DIFF
--- a/src/components/ArrowLink/ArrowLink.tsx
+++ b/src/components/ArrowLink/ArrowLink.tsx
@@ -13,6 +13,7 @@ type ArrowLinkProps = {
 	replace?: boolean;
 	className?: string;
 	back?: boolean;
+	isLinkToExternalUrl?: boolean;
 };
 
 export const TEST_ID = {
@@ -30,6 +31,7 @@ const ArrowLink: React.FC<ArrowLinkProps> = ({
 	replace,
 	style,
 	back,
+	isLinkToExternalUrl,
 }) => {
 	const content = (
 		<>
@@ -58,7 +60,7 @@ const ArrowLink: React.FC<ArrowLinkProps> = ({
 	 * refactor this in the future.
 	 */
 	return href ? (
-		<Link {...sharedProps} to={href}>
+		<Link {...sharedProps} to={isLinkToExternalUrl ? { pathname: href } : href}>
 			{content}
 		</Link>
 	) : (
--- a/src/containers/other/NFTView/elements/TokenHashBoxes/TokenHashBoxes.tsx
+++ b/src/containers/other/NFTView/elements/TokenHashBoxes/TokenHashBoxes.tsx
@@ -83,6 +83,7 @@ export const TokenHashBoxes: React.FC<TokenHashBoxesProps> = ({
 						width: 150,
 					}}
 					href={etherscanLink}
+					isLinkToExternalUrl
 				>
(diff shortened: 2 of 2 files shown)


src/components/ArrowLink/ArrowLink.tsx
```diff
@@ -58,7 +60,7 @@ const ArrowLink: React.FC<ArrowLinkProps> = ({
 	 * refactor this in the future.
 	 */
 	return href ? (
-		<Link {...sharedProps} to={href}>
+		<Link {...sharedProps} to={isLinkToExternalUrl ? { pathname: href } : href}>
 			{content}
 		</Link>
```
colbr, line 65: The way this was originally set up was so that `to` would refer to an internal location (e.g. `/market`), and `href` would refer to an external location, mainly to mimic the API difference between `<Link>` and `<a>` (one requiring `to` and the other requiring `href`)  - I think that regressed at some point, but it would have looked something like like:

```
if(!href && !to) return;
else return <Link to={ href ? { pathname: href } : to } />
```

or maybe the second line could have been:

```
href ? <a href={href} /> : <Link to={to} />
```

It was originally designed this way so that if you needed an internal link (`<Link>`), you could pass `ArrowLink` the usual prop that you'd use for a `<Link>` (`to`), and if you wanted an external link (`<a>`) you could use the usual `href`. I'm realising that this design decision may have made the API a little more confusing than it should be, as the distinction between `to` and `href` isn't obvious!

Not really sure what's going on inside the `<Link>` component, but would it be advantageous to just render an `<a>` when `isLinkToExternalUrl === true`? This way it would skip whatever happens internally in `<Link>`, which I assume is just logic for navigating around a SPA (current URL context, etc).

colbr (approved):
Nice! One smol question re `ArrowLink`, but not needed for approval

src/components/ArrowLink/ArrowLink.tsx
domw30: Brett thanks for the write up there. Yes I did some digging on this and totally agree that for external links we should render an `<a>`. I found that `<Link>` should really only be used for internal react routing. 

There was a comment left above the condition for returning a `<Link>` or `<a>` stating something about 'still rendering `<a>` tags', so I diverted away from that thinking there was a reason for not using `<a>`, seemed unwanted 😆 

I'll note to update this

eric-zero-dev: I recommend to customize `Link` component to accept both of `to` and `href` for internal and external link.

So if there is `to` in props, it will be react-router link(internal) and if there is `href` it will be `<a>` tag rendering.

How do you think guys? @colbr @domw30

domw30: Hey Eric, sorry for delay. Yeah this makes sense to me. I'll aim to update this during the week :)

PR #546, opened 2022-05-02
fix(profile): profile re-rendering on hover

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Hovering-over-network-wallet-button-causes-to-many-requests-fb0411f7566b425cbf5550ab9b85b31b)

-----

## 1. Pull request checklist

- [x]  A reviewer has been assigned to the Notion card

## 2. PR type
 

Bugfix

Hovering over the account widget in Profile would cause tables to re-rendering, causing a cascade of network requests and repaints.

Hovering over the account widget has no effect on the other Profile content.

## 5. Other information

Memoise content so it doesn't re-render when unrelated Profile state changes - in this case state related to the account widget. In the next iteration, we could make the account widget its own component.


colbr:
Approved by @EthFish in Notion

PR #553, opened 2022-05-03
Fix: 0.15.0 (Reskin) Bundle of Bugs/Updates

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=b82bd575da3e48b29a0753ab410e4549)

-----

## 1. Pull request checklist

- [x]  Notion card has a link to this PR

## 2. PR type
Mixture of bugfixes/hotfixes and design changes

Place a Bid - No ‘$’ Before USD value
Buy Now - Sub title message should not display when processing transaction
Possible to accept an already ‘accepted’ bid
Staking flow shows two headers
Connect Wallet Button is too big in desktop screen
Update useAcceptBid tests and resolve failing tests
Staking Modal should not have a header
Stat card in NFT view should be same height as other cards
Domain settings - cross icon out of position
Domain settings - domain bidding and grid view by default should be toggled by default

The above issues have been resolved.

### DIFF
--- a/src/components/NFTMedia/NFTMedia.module.scss
+++ b/src/components/NFTMedia/NFTMedia.module.scss
@@ -6,7 +6,6 @@
 	height: 100%;
 	width: 100%;
 	overflow: hidden;
-	border-radius: 8px 8px 0 0;
 }
 
 .Media {
@@ -14,7 +13,6 @@
 	max-width: 100%;
 	max-height: 100%;
 	transition: opacity var(--animation-time-medium) ease-in-out;
-	border-radius: 8px 8px 0 0;
 }
 
 .Cover {
--- a/src/components/NFTMedia/NFTMedia.module.scss
+++ b/src/components/NFTMedia/NFTMedia.module.scss
@@ -6,6 +6,7 @@
 	height: 100%;
 	width: 100%;
 	overflow: hidden;
+	border-radius: 8px 8px 0 0;
 }
 
 .Media {
@@ -13,6 +14,7 @@
 	max-width: 100%;
 	max-height: 100%;
 	transition: opacity var(--animation-time-medium) ease-in-out;
+	border-radius: 8px 8px 0 0;
 }
 
 .Cover {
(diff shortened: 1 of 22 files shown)


src/components/NFTMedia/NFTMedia.module.scss
```diff
@@ -6,13 +6,15 @@
 	height: 100%;
 	width: 100%;
 	overflow: hidden;
+	border-radius: 8px 8px 0 0;
```
colbr, line 9: `NFTMedia` is used in a few different places, some of which don't have `border-radius` (`NFTView` for example). Can we set `border-radius` in the component rendering the `NFTMedia` by passing a `className` (or similar)?

src/containers/other/NFTView/elements/DomainSettings/elements/DomainSettingsBody/hooks/useDomainSettingsBodyData.tsx
```diff
@@ -18,8 +18,8 @@ export const useDomainSettingsBodyData = (
 			'';
 		const story = metadata?.description || '';
 		const isMintable = Boolean(metadata?.isMintable);
-		const isBiddable = Boolean(metadata?.isBiddable);
-		const gridViewByDefault = Boolean(metadata?.gridViewByDefault);
+		const isBiddable = true;
+		const gridViewByDefault = true;
```
colbr, line 22: This will break if metadata has `isBiddable: false` or `gridViewByDefault: false` in metadata. These lines should be able to stay as is, and defaults should be set in `metadata.ts:parseMetadata`

colbr (changes requested):
Looking awesome, you've got a great eye for detail.

Two small changes 🚀

src/components/NFTMedia/NFTMedia.module.scss
domw30: Hmm I checked the images again and looked as though these radius' weren't required. I've just removed these, going to let it build a preview then check out on mobile.

src/containers/other/NFTView/elements/DomainSettings/elements/DomainSettingsBody/hooks/useDomainSettingsBodyData.tsx
domw30: Makes sense thanks. Reverted this change however, the settings won't be toggled as these return false.

colbr (approved):
LGTM 🚀

PR #566, opened 2022-05-04
add maintenance page

padala:

-----

## 1. Pull request checklist


eric-zero-dev (approved):
Looks good to me. I have an idea to switch maintanance mode and normal mode. But since it is urgent thing, lets do it for now.

padala:
> Looks good to me. I have an idea to switch maintanance mode and normal mode. But sine it is urgent thing, lets do it for now.

Yeah, we should have a concrete solution for this. Both from a code and infrastructure standpoint. We are caught off-guard on this.

eric-zero-dev:
> > Looks good to me. I have an idea to switch maintanance mode and normal mode. But sine it is urgent thing, lets do it for now.
> 
> Yeah, we should have a concrete solution for this. Both from a code and infrastructure standpoint. We are caught off-guard on this.

Can you please create a card and assign me? I think we can handle it without code change in netlify

PR #570, opened 2022-05-11
fix: loading spinner below transparent image

eric-zero-dev:
[Associated Notion Card](https://www.notion.so/zerotech/Fix-spinner-in-background-of-transparent-images-in-ImageCard-c94beb4d6e544f32bbf4769d3db5ca08)

-----

## 1. Pull request checklist

## 2. PR type
 
Bugfix

### DIFF
--- a/src/components/Cards/ImageCard/ImageCard.module.scss
+++ b/src/components/Cards/ImageCard/ImageCard.module.scss
@@ -1,6 +1,7 @@
 @use 'styles/variables/dimensions';
 
 $placeholder-background: #000;
+$media-background: #000;
 
 .Container {
 	--image-card-padding: 100%;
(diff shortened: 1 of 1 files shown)


src/components/Cards/ImageCard/ImageCard.module.scss
```diff
@@ -1,6 +1,7 @@
 @use 'styles/variables/dimensions';
 
 $placeholder-background: #000;
+$media-background: #000;
```
colbr, line 4: One small nit:

Maybe we could use `$placeholder-background` in both places rather than assigning two variables, or we could set `$media-background: $placeholder-background` if you kept them separate to be more explicit.

colbr (approved):
Huh, this is clever. Much better solution than I was thinking of - nice work!

PR #573, opened 2022-05-11
refactor: variables refactor / improvements & pay off tech debt

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Projects-e96437225e264ae8ae8d46ca5f4d7b35?p=174fa9aad25a4685ad6568214507584f)

-----

## 1. Pull request checklist

## 2. PR type
- code style
- refactor

Main style file required a tidy up and instead use style variables file.
Some tech debt from previous release.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,4 +1,4 @@
-@use './variables/colors.scss';
+@use './variables/colors';
 @use './variables/dimensions';
 @use './variables/animations';
 @use './variables/typography';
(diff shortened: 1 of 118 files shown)


src/styles/main.scss
```diff
@@ -13,92 +13,7 @@
  */
 
 :root {
-	--font-size-default: #{typography.$size-default};
```
colbr, line 16: Satisfying block of removals 🔥

src/styles/variables/_colors.scss
```diff
@@ -1,9 +1,109 @@
+// Black
+$off-black-9: rgba(0, 0, 0, 0.9);
+$off-black-73: rgba(0, 0, 0, 0.73);
+$off-black-5: rgba(0, 0, 0, 0.5);
+$off-black-3: rgba(0, 0, 0, 0.3);
 $black: #000000;
```
colbr, line 6: Could be slightly better design to not include opacity variants in variables, and instead use `rgba($black, [opacity])` when we use an off-black value. Otherwise, we might run into the situation where we have a long list of variables for arbitrary opacity amounts.

src/styles/variables/_colors.scss
```diff
@@ -1,9 +1,109 @@
...
+// Blues
+$anakiwa: #a1d2ff;
+$anakiwa-darker: #88d7ff;
+$anakiwa-lighter: #9ed0ff;
+$baby-blue: #9dd0ff;
+$blumine-darker: rgba(30, 81, 128, 0.2);
+$doger-blue-lighter: #3ca1ff;
+$doger-blue-darker-2: rgba(60, 161, 255, 0.2);
+$doger-blue-darker-05: rgba(60, 161, 255, 0.05);
+$doger-blue-darker-08: rgba(60, 161, 255, 0.08);
```
colbr, line 17: Similar to the previous comment - could be better to handle opacity variants when they're used rather than explicitly setting them in vars.

src/styles/variables/_colors.scss
```diff
@@ -1,9 +1,109 @@
+// Black
```
colbr, line 1: Do these variable names line up with what's in Figma at the moment? I think the ultimate DX here would be if these variables matched the Figma colour names, so you don't have to look up what we call it in our code base.

colbr (commented):
Looking nice - a question or two in the comments before I hit the green button. Mainly wondering how we should handle opacity variants of colours, and if it would be possible to map our colour names to the Figma colour names for better DX.

domw30:
hey @colbr. I re-did the colour variables and aligned them with whats in figma. There are some colours that aren't named in the designed so I left them as they were for now, but most colours should now have improved the DX :) 

Ready for another re-view

colbr (approved):
Nice work 🔥

PR #576, opened 2022-05-16
feat: DAO proposal detail view

eric-zero-dev:
[Associated Notion Card](https://www.notion.so/zerotech/Implement-UI-ec0029ca125d4fc3af7d8507cf9ba43f)

-----

## 1. Pull request checklist

## 2. PR type
 
Feature

I created some structure of proposal lists and detail page in `feat/dao-proposals` branch. We had to implement UI for proposals and detail page

Implemented of proposal detail page


src/containers/DAO/pages/DAOPage/Proposals/ProposalDetail/VoteAttributes/VoteAttributes.tsx
```diff
@@ -0,0 +1,169 @@
...
+		return result;
+	}, [proposal, metadata]);
+
+	const initialHiddenAttributesCount: number = useMemo(() => {
+		if (attributes.length > initialVisibleAttributesCount) {
+			return attributes.length - initialVisibleAttributesCount;
+		}
+
+		return 0;
+	}, [attributes, initialVisibleAttributesCount]);
```
colbr, line 120: I think `useMemo` might be less efficient here, compared to calculating the value per render. `useMemo` has some background overhead from having to cache a value and recall it, so in situations where the calculation isn't resource intensive it doesn't provide a performance boost (sometimes the opposite).

Could instead be
```js
return Math.max(attributes.length - initialVisibleAttributesCount, 0)
```

src/containers/DAO/pages/DAOPage/Proposals/ProposalDetail/VoteBar/VoteBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+					className={`${styles.ProgressbarItem} ${styles.ProgressbarItemRight}`}
+					style={{ width: `${percentage2}%` }}
+				></div>
+			</div>
+			<span className={`${styles.ValueLabel} ${styles.ValueLabelRight}`}>
+				{percentage2}%
+			</span>
+		</div>
+	);
+};
```
colbr, line 35: Nice!

src/lib/utils/string.ts
```diff
@@ -13,3 +15,10 @@ export const truncateString = (
 	str.length > len
 		? str.slice(0, len > tail.length ? len - tail.length : len) + tail
 		: str;
+
+/**
+ * Capitalize string
+ * @param str to capitalize
+ * @returns capitalized string
+ */
+export const capitalizeString = (str: string) => capitalize(str);
```
colbr, line 24: Is it worth re-defining/re-exporting the lodash `capitalize` method here? I think it makes sense to do this when you're expanding an external library function, or when you're not 100% sold on the library function and would like to easily swap it out at some point. In this case, `capitalize` is a very standard function and won't be extended or replaced in the future - may be worth just using `capitalize` straight from lodash to reduce code base bloat.

colbr (commented):
Looking awesome so far! Left a couple of comments, I also noticed there's a couple things you're still working on so will wait for it all to be done before final review. Nice work! 🚀

colbr (approved):
Looking good!

PR #578, opened 2022-05-19
 fix: table iteration feedback

joonchoi0918:
[Associated Notion Card](https://www.notion.so/zerotech/Table-Feedback-16bdb553718441368d97637a98a40394)

-----

## 1. Pull request checklist

## 2. PR type
 
BugFix

1. Proposals table refreshed whenever the `Proposals` tab clicked
2. Wrong font size, color, margin, ..etc in Proposals Table

### DIFF
--- a/src/components/Chiclet/Chiclet.module.scss
+++ b/src/components/Chiclet/Chiclet.module.scss
@@ -4,6 +4,8 @@
 	align-items: center;
 	padding: 8px 16px;
 	font-size: 12px;
+	font-weight: 700;
+	line-height: 15px;
 	color: #D9D9D9;
 	border: 1px solid #3B3B3B;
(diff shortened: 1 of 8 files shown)


colbr (changes requested):
Looking good!

I can't remember if I mentioned it in the Loom, but the `Status` on the list view and the grid view are wrong. If the proposal voting has ended, the status should be the outcome of the vote - i.e. "vote passed", "vote failed" (whatever it is in Figma). If proposal voting is still open, it should say what option is leading, e.g. "approval favoured", "denial favoured" (whatever it is in Figma)

Other than that, looks perfect 🔥

colbr (approved):
Going to merge this now as it has to go in, and we can fix the feedback later

PR #579, opened 2022-05-19
fix: proposals detail page iteration feedback

eric-zero-dev:
[Associated Notion Card](https://www.notion.so/zerotech/Page-Proposal-Detail-Page-aka-Voting-Page-eb7247ea0f2942b881362c9b4dcc19bb)

-----

## 1. Pull request checklist

## 2. PR type
 
Bugfix

1. There was included DAO navigation
2. Font color & number format was wrong


colbr (changes requested):
Nice, this is looking much better.

It looks like you missed a few things from my feedback:
- Remove padding from markdown viewer component (should line up with the edges)
- Code viewer in markdown is currently white text on white background - should be white text on a dark background
- Votes submitted always shows "-", even when there are votes
There's a few other things that I can't think of off the top of my head, but make sure to go back through the video and jot down everything I mentioned.

Finish those off and we can merge 🚀

src/containers/DAO/pages/DAOPage/hooks/useProposal.ts
```diff
@@ -1,34 +1,76 @@
...
-	useEffect(() => {
-		setProposal(undefined);
-		setIsLoading(true);
-		dao
-			?.getProposal(id)
-			.then((p) => {
-				setProposal(p);
-			})
-			.catch((e) => {
+	const fetchVotes = useCallback(
```
colbr, line 21: I don't think we need `useCallback` here (same with the one below)

colbr (changes requested):
Looking good, nice work. Just leaving comment for now as I know you're working on another couple things for this PR!

colbr (approved):

PR #591, opened 2022-06-06
fix(place-bid): update place bid to use spend token approval instead of nft transfer

padala:
[Incorrect approval in MakeABid](https://www.notion.so/zerotech/Incorrect-approval-in-MakeABid-ce320e8a6c55431a841b9ca46a11eacf)

-----

## 1. Pull request checklist


src/containers/flows/MakeABid/MakeABid.tsx
```diff
@@ -107,13 +107,15 @@ const MakeABid = ({ domain, onBid, onClose }: MakeABidProps) => {
...
 			try {
-				const zAuction = await sdk.getZAuctionInstanceForDomain(domain.id);
-				const isApproved = await zAuction.isZAuctionApprovedToTransferNft(
-					account,
-				);
+				const needsApproval =
+					await sdk.zauction.needsToApproveZAuctionToSpendTokens(
+						domain.id,
+						account,
+						BigNumber.from('1000000000'),
```
colbr, line 114: Think this should be changed in the future as it's not uncommon to send billions of a token if the token price is very low. All good for now though as we're just using WILD and Zero.

colbr (approved):
Nice work! 🚀

Can you make a card in Notion to change the `needsToApproveZAuctionToSpendTokens` amount to `bid amount` in the future? It will require some design changes (check will need to happen after user enters bid amount) so it's out of scope for this one.

Side note, you don't need to worry about bumping version number when going `feature branch -> develop` - we just do that when going `develop -> master`.

src/containers/flows/MakeABid/MakeABid.tsx
padala: We don't necessarily have to modify the flow Zach said you can just check for a really large approval like a billion because we always approve in the SDK for `Number.MAX_SAFE_INTEGER` anyways

PR #593, opened 2022-06-10
feat: Maintenance page for dApp (DomW)

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Team-Dash-78b824c0cd0e46c59c19dfe103ae929a?p=4c2a7bc177bb4c7db8842416fd5f904a)

-----

## 1. Pull request checklist

## 2. PR type
- [x] Feature

Maintenance page did not exist, we were only displaying page content to reflect a maintenance page.
Only one form of maintenance content had been built.

Maintenance page has been created and added, implemented with page container and logic added to determine which page container to use.
Types of maintenance added.


domw30:
Hey @padala, would it be possible to receive some feedback around the approach for maintenance pages please. I'm wondering if I should create 3 separate pages rather than changing page content, would you have any thoughts here? Placing the maintenance route in app.tsx doesn't have the desired outcome at the moment, the content loads on each page.

PR #604, opened 2022-06-27
feat: switch network based on .env

colbr:
_Filling this out shortly_


src/components/SideBar/SideBar.tsx
```diff
@@ -75,16 +98,17 @@ const SideBar = () => {
...
-						>
-							<img alt={ALT_TEXT.ZERO_LOGO} src={DOMAIN_LOGOS.ZERO} />
-						</a>
+					<PriceWidget isRoot={!zna.startsWith('wilder')} />
+					<div
+						className={cx(styles.ZeroIconContainer, {
+							Hidden: isDefaultNetworkRootPath,
+						})}
+					>
+						{zna.startsWith('wilder') && (
```
domw30, line 107: Ah I can see why you've removed the zero.tech link as we'd be displaying two zero logos in the sidebar which would look a bit weird. 

This does change the behaviour though, if a user clicks into 'tester' network, the zero logo won't take them to the zero root page, so this could be a bit misleading. Something we can pick up in due course

src/lib/providers/CurrentDomainProvider.tsx
```diff
@@ -22,42 +25,45 @@ export const CurrentDomainContext = React.createContext({
...
+		document.title = process.env.REACT_APP_TITLE + ' | Market';
+	}
+
 	const contextValue = {
 		domain: znsDomain.domain,
 		domainId,
 		domainRaw: domain,
 		domainMetadata,
-		app: location.pathname.indexOf('/market') > -1 ? '/market' : '/staking',
+		app: appFromPathname(pathname),
```
domw30, line 66: I'm going to revert this to the change I made previously on my branch. 

```
pathname.indexOf(ROUTES.MARKET) > -1
		? ROUTES.MARKET
		: ROUTES.STAKING + ROUTES.STAKING_POOLS,
```

domw30 (approved):

PR #605, opened 2022-06-27
fix: Error Handling (flows)

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Team-Dash-78b824c0cd0e46c59c19dfe103ae929a?p=3a00de36dfad480e957d4afd630ec2d5)

-----

## 1. Pull request checklist

## 2. PR type
    - Bugfix

Accept bid error not specific enough. Text would always read 'Reject by wallet' for any error.
SetBuyNow errors not specific enough
BuyNow errors not specific enough
CancelBid errors not specific enough
MakeABid errors not specific enough
TransferOwnership errors not specific enough

Accept bid error handling improved. Generic error message added and some logic to determine generic or rejected message.
SetBuyNow specific errors
BuyNow  specific errors
CancelBid specific errorsh
MakeABid  specific errors
TransferOwnership specific errors
Error handling across wallet types tested (metamask & coinabase wallet)
Updated tests to reflect changes where required

### DIFF
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -1,5 +1,5 @@
 export const ERRORS = {
-	REJECTED_WALLET: 'Rejected by wallet.',
+	REJECTED_WALLET: 'Rejected by wallet. Please try again.',
 	DATA_CONSUMED:
 		'The bid you are trying to accept is no longer valid. Please select another bid to accept',
 	PROBLEM_OCCURRED: 'An unexpected problem has occurred. Please try again.',
@@ -9,6 +9,7 @@ export const ERRORS = {
 	LIBRARY_NOT_FOUND: 'Could not find web3 library',
 	FAILED_TO_CHECK_ZAUCTION: 'Failed to check zAuction approval status',
 	FAILED_TO_LOAD_DOMAIN_ID: 'Failed to load domain ID',
+	FAILED_TO_RETRIEVE_BID_DATA: 'Failed to retrieve bid data',
 	FAILED_TO_GET_DATA: 'Failed to Get Data',
 };
 
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -11,6 +11,8 @@ export const ERRORS = {
 	FAILED_TO_LOAD_DOMAIN_ID: 'Failed to load domain ID',
 	FAILED_TO_RETRIEVE_BID_DATA: 'Failed to retrieve bid data',
 	FAILED_TO_GET_DATA: 'Failed to Get Data',
+	INSUFFICIENT_FUNDS_BID:
+		'You don’t have enough WILD to make that large of a bid.',
 };
 
 export const MESSAGES = {
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -8,6 +8,7 @@ export const ERRORS = {
 	LIBRARY: 'Failed to connect with Web3 wallet.',
 	LIBRARY_NOT_FOUND: 'Could not find web3 library',
 	CONSOLE_TEXT: 'Failed to check zAuction approval status',
+	FAILED_TO_LOAD_DOMAIN_ID: 'Failed to load domain ID',
 };
 
 export const MESSAGES = {
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -11,8 +11,6 @@ export const ERRORS = {
 	FAILED_TO_LOAD_DOMAIN_ID: 'Failed to load domain ID',
 	FAILED_TO_RETRIEVE_BID_DATA: 'Failed to retrieve bid data',
 	FAILED_TO_GET_DATA: 'Failed to Get Data',
-	INSUFFICIENT_FUNDS_BID:
-		'You don’t have enough WILD to make that large of a bid.',
 };
 
 export const MESSAGES = {
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -15,5 +15,6 @@ export const ERRORS = {
 
 export const MESSAGES = {
 	TRANSACTION_DENIED: 'denied transaction signature',
+	MESSAGE_DENIED: 'denied message signature',
 	DATA_CONSUMED: 'data already consumed',
 };
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -19,4 +19,5 @@ export const MESSAGES = {
 	TRANSACTION_DENIED: 'denied transaction signature',
 	MESSAGE_DENIED: 'denied message signature',
 	DATA_CONSUMED: 'data already consumed',
+	FAILED_TO_SIGN_BID: 'Failed to sign bid message',
 };
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -1,0 +1,15 @@
+export const ERRORS = {
+	REJECTED_WALLET: 'Rejected by wallet.',
+	DATA_CONSUMED:
+		'The bid you are trying to accept is no longer valid. Please select another bid to accept',
+	PROBLEM_OCCURRED: 'An unexpected problem has occurred. Please try again.',
+	SIGNATURE: 'Failed to generate signature.',
+	TRANSACTION: 'Transaction failed. Please try again',
+	LIBRARY: 'Failed to connect with Web3 wallet.',
+	CONSOLE_TEXT: 'Failed to check zAuction approval status',
+};
+
+export const MESSAGES = {
+	TRANSACTION_DENIED: 'denied transaction signature',
+	DATA_CONSUMED: 'data already consumed',
+};
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -6,6 +6,7 @@ export const ERRORS = {
 	SIGNATURE: 'Failed to generate signature.',
 	TRANSACTION: 'Transaction failed. Please try again',
 	LIBRARY: 'Failed to connect with Web3 wallet.',
+	LIBRARY_NOT_FOUND: 'Could not find web3 library',
 	CONSOLE_TEXT: 'Failed to check zAuction approval status',
 };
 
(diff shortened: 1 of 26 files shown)


domw30:
@colbr, noticed set buy now error handling could be improved. Let me know if you want me to add on this PR.

colbr:
> @colbr, noticed set buy now error handling could be improved. Let me know if you want me to add on this PR.

That would be great - update the Notion card too. Thanks!

domw30:
@colbr, ready for review :)

src/constants/errors.ts
```diff
@@ -0,0 +1,23 @@
+export const ERRORS = {
```
domw30, line 1: Decided to start extracting potential error messages into its own file as in some cases we use the same messaging

src/containers/flows/CancelBid/CancelBid.tsx
```diff
@@ -41,13 +41,13 @@ export const CancelBid = ({
 
 	const onFinish = () => {
 		onClose();
+		onSuccess();
```
colbr, line 44: If `onSuccess` is here, it won't run if the user closes the modal, or if the user waits until the `tx` finishes and closes the modal by clicking out of it, rather than hitting the "Finish" button. Not sure if this is the intended behaviour, but worth considering.

src/containers/flows/CancelBid/hooks/useCancelBid.ts
```diff
@@ -52,9 +56,10 @@ const useCancelBid = (): UseCancelBidReturn => {
 					cancelBidOnChain,
 					library.getSigner(),
 				);
-			} catch (e) {
+			} catch (e: any) {
 				console.error(e);
-				throw new Error(constants.ERRORS.SIGNATURE);
+				const errorText = getDisplayErrorMessage(e.message);
+				throw new Error(errorText);
```
colbr, line 62: nit: I generally like to just do `throw new Error(getDisplayErrorMessage(e.message))` rather than assign a variable and then do a single operation with it.

src/containers/flows/SetBuyNow/index.tsx
```diff
@@ -133,20 +156,21 @@ const SetBuyNowContainer = ({
...
 					onSuccess();
 				}
-			} catch (e) {
+			} catch (e: any) {
 				setCurrentStep(Step.SetBuyNow);
-				setError((e as any).message);
-				console.log('Error setting buy now price', e);
+				console.error(e);
+				const errorText = getDisplayErrorMessage(e.message);
+				setError(errorText);
```
colbr, line 173: nit: similar to above with `setError(getDisplayErrorMessage(e.message))`

src/containers/flows/SetBuyNow/index.tsx
```diff
@@ -120,11 +132,22 @@ const SetBuyNowContainer = ({
...
+				try {
+					setCurrentStep(Step.SettingBuyNow);
+					await tx?.wait();
+					// Timeout to prevent jolt
+					if (!amount) {
+						await new Promise((r) => setTimeout(r, 500));
+					}
+				} catch (e) {
+					setCurrentStep(Step.SetBuyNow);
+					setError(ERRORS.TRANSACTION);
```
colbr, line 148: Same issue as earlier with continued execution after the `catch`. `return` or `throw` from `catch`.

src/lib/hooks/useBidProvider.tsx
```diff
@@ -202,11 +203,16 @@ export const useBidProvider = (): UseBidProviderReturn => {
...
 			if (!library) {
-				console.error('Could not find web3 library');
+				console.error(ERRORS.LIBRARY_NOT_FOUND);
+				return;
+			}
+
+			if (!domain.id) {
+				console.error(ERRORS.FAILED_TO_LOAD_DOMAIN_ID);
 				return;
 			}
```
colbr, line 217: Could be worth `throw`ing these errors to the parent `catch` - this works as well though.

src/lib/hooks/useTransfer.tsx
```diff
@@ -40,32 +41,46 @@ export const useTransfer = (): UseTransferReturn => {
...
+				// Signature request
+				try {
+					tx = await sdk.transferDomainOwnership(
+						params.walletAddress,
+						params.domainId,
+						library.getSigner(),
+					);
+				} catch (e: any) {
+					console.error(e);
+					throw new Error(e.message);
```
colbr, line 68: Nice `throw` 🔥

colbr (changes requested):
Few changes related to `try...catch` - have a wee run around in codesandbox.io to check behaviour of `try...catch`. Looking nice though, thanks for picking this up.

padala:
@domw30  Probably a good idea to update the base branch to zAuction and merge it back into this to avoid lot of conflicts at a later stage.

src/containers/flows/BuyNow/BuyNow.tsx
```diff
@@ -73,6 +74,17 @@ const BuyNow = ({
...
+	if (step === Step.FailedToCheckZAuction) {
+		return WrapWizard(
+			<Wizard.Confirmation
+				error={error}
+				message={'Failed to check zAuction approval status'}
+				primaryButtonText={'Close'}
+				onClickPrimaryButton={onCancel}
+			/>,
+		);
+	}
```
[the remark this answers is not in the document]
domw30: Okay I updated this to include an extra ```setCurrentStep``` when failing zAuction approval. From memory, this will need to be updated on other flows as well. I think having the 'close' button and failed zAuction approval step were initially by design but it does make sense to include a retry option, rather than close the modal and re-open the flow. The update I've made is shown below. Let me know what you think, works for me, could be one for design though..

https://user-images.githubusercontent.com/39112648/177580147-7a05b3b3-6788-4ac7-8404-5c192b6faec1.mov

src/constants/errors.ts
```diff
@@ -0,0 +1,23 @@
...
+	SIGNATURE: 'Failed to generate signature.',
+	TRANSACTION: 'Transaction failed. Please try again',
+	LIBRARY: 'Failed to connect with Web3 wallet.',
+	LIBRARY_NOT_FOUND: 'Could not find web3 library',
+	FAILED_TO_CHECK_ZAUCTION: 'Failed to check zAuction approval status',
+	FAILED_TO_LOAD_DOMAIN_ID: 'Failed to load domain ID',
+	FAILED_TO_RETRIEVE_BID_DATA: 'Failed to retrieve bid data',
+	FAILED_TO_GET_DATA: 'Failed to Get Data',
+	INSUFFICIENT_FUNDS_BID:
+		'You don’t have enough WILD to make that large of a bid.',
```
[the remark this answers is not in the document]
domw30: Agreed, I think the only one here I'd expect not to see reused elsewhere in is INSUFFICIENT_FUNDS_BID - I'll move it. The intention here is when we get round to tidying up some of the other flows (removing hard code) we should be able to re-use these constants. It makes sense to re-use the same error copy throughout the dApps, at the moment there are minor variations of the above, just need to go in and replace. I would have done this with the mind set that if its used more than once it can be global const....I think!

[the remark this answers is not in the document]
domw30: Done

src/containers/flows/CancelBid/CancelBid.tsx
domw30: Ah yeah that makes sense. I changed it to this to prevent the modal closing once ```tx?.wait()``` was complete in ```useCancelBid```. I've reverted it now but unsure how to prevent the modal closing and therefore skipping the Success step of the flow. Would you have any ideas why the modal is closing as below?

https://user-images.githubusercontent.com/39112648/178272796-d0105b1d-8221-46f1-b3ad-f51a10778fef.mov

src/containers/flows/CancelBid/hooks/useCancelBid.ts
domw30: Yeah sounds good to me, not sure why but I thought that was the practice to follow but never made sense to me having the additional line, unless being reused. Thanks. Updated

src/containers/flows/BuyNow/index.tsx
```diff
@@ -95,22 +103,28 @@ const BuyNowContainer = ({
...
-			setCurrentStep(Step.Buying);
-			await tx.wait();
-			addNotification(`You have successfully purchased ${data?.title}`);
+			try {
+				setCurrentStep(Step.Buying);
+				await tx.wait();
+			} catch (e) {
+				setCurrentStep(Step.Details);
+				setError(ERRORS.TRANSACTION);
+			}
```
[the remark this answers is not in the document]
domw30: Thanks for that, does the change here make sense to pass ERRORS.TRANSACTION? My understanding is this would throw the transaction failed message to the above catch, resulting in the error printing in the console, and the user would see 'Unexpected problem .....' (generic message), because I'm using the getDisplayErrorMessage function in the above catch

```setError(getDisplayErrorMessage(e.message));```

File: src/containers/flows/CancelBid/CancelBid.tsx (hunk 2 above)
Comment (colbr) [reply src/containers/flows/CancelBid/CancelBid.tsx]: Ahhh I see - the `onSuccess` will be triggering an API call to get your bids, and since you just cancelled your bid, the button won't render and therefore the modal will disappear.

Could use an `isMounted` ref, and after `tx.wait()` check if `isMounted.current === false`. If yes, call `onSuccess()`, otherwise show success screen.

File: src/containers/flows/CancelBid/hooks/useCancelBid.ts (hunk 3 above)
Comment (colbr) [reply src/containers/flows/CancelBid/hooks/useCancelBid.ts]: I think it makes sense when it's not clear what the variable is - for function calls with one parameter, or when using functions/variables with already explicit names, it's good practice.

**Review by colbr — CHANGES_REQUESTED** [review]
Looking good, but I did notice some fallthrough cases for `try...catch` blocks.

Remember if you
```js
try {
    throw;
} catch {
    // handle error
}
console.log('hello')
```
Console will log "hello".

Specifically referencing `MakeABid:152` - if the approval `tx.wait` fails it will run the catch block then continue with line 59 and line 60. In this example, line 59 and 60 need to be moved to the try  block. Not sure why I couldn't select those lines for review!

PR #607, opened 2022-07-01
feat: Create dao proposal

joonchoi0918:
[Associated Notion Card](https://www.notion.so/zerotech/UI-SDK-Implementation-61d3883d63894326a63dd2c41075e786)

-----

## 1. Pull request checklist

## 2. PR type
 
    - Feature

There was Dao Proposals MVP with proposals list

Added Create New Proposal page


src/containers/DAO/pages/DAOPage/Proposals/CreateProposal/CreateProposalNotes.tsx
```diff
@@ -0,0 +1,73 @@
...
+			{/* Connect Wallet Note */}
+			{notes.ConnectWallet.show && (
+				<div className={styles.Notes_Section}>
+					<div className={styles.Notes_Section_Note}>
+						Please connect a Web3 wallet to create a proposal
+					</div>
+					<div className={styles.Notes_Section_Action}>
+						<FutureButton glow onClick={notes.ConnectWallet.onClick}>
+							Connect Wallet
+						</FutureButton>
```
colbr, line 36: Use `ConnectWalletButton` here instead.

Also just wondering - is there a reason you've handled the show/hide "Connect Wallet" modal logic in the hook and passed values as props into this component, instead of just handling the state in here? I think it's over-complicating things here.

src/containers/DAO/pages/DAOPage/Proposals/CreateProposal/hooks/useCreateProposal.ts
```diff
@@ -0,0 +1,89 @@
...
+			nav.style.display = 'none';
+		}
+	});
+
+	useWillUnmount(() => {
+		const nav = document.getElementById('dao-page-nav-tabs');
+		if (nav) {
+			nav.style.display = 'block';
+		}
+	});
```
colbr, line 30: nit: could be simplified

```js
useEffect(() => {
    const nav = document.getElementById('dao-page-nav-tabs');
    if(nav) {
        nav.style.display = 'none';
    }
   return () => {
        nav.style.display = 'block';
    }
}, [])
```

I think this functionality might be outside of the scope of this hook though, this hook is doing a lot of stuff.

src/containers/DAO/pages/DAOPage/Proposals/CreateProposal/hooks/useCreateProposalForm.ts
```diff
@@ -0,0 +1,319 @@
...
+} from '../CreateProposal.constants';
+import {
+	getTokenOption,
+	getVotingDetails,
+	isValidERC20Address,
+	validateCreateProposalForm,
+} from '../CreateProposal.helpers';
+import { DAO_CREATE_PROPPAL } from '../../../Proposals/Proposals.constants';
+
+export const useCreateProposalForm = ({
```
colbr, line 30: This hook is doing too much. Hooks should have a clearly defined purpose - at the moment this is driving the entirety of the UI and handling the business logic.

Can we break this down to be simpler?

src/containers/DAO/pages/DAOPage/Proposals/CreateProposal/hooks/useCreateProposal.ts
```diff
@@ -0,0 +1,89 @@
...
+import { useWeb3React } from '@web3-react/core';
+import { Web3Provider } from '@ethersproject/providers/lib/web3-provider';
+import type { zDAO } from '@zero-tech/zdao-sdk';
+import { useDidMount } from 'lib/hooks/useDidMount';
+import { useWillUnmount } from 'lib/hooks/useWillUnmount';
+import { getTokenOptionsFromAssets } from '../CreateProposal.helpers';
+import { DAO_CREATE_PROPPAL } from '../../../Proposals/Proposals.constants';
+import useAssets from '../../../hooks/useAssets';
+
+export const useCreateProposal = (dao?: zDAO) => {
```
colbr, line 12: I think this hook is doing a lot of stuff it shouldn't be responsible for, such as handling navigation, component lifecycle, and "Connect Wallet" modal state.

src/containers/DAO/pages/DAOPage/Proposals/ProposalDetail/ProposalAttributes/ProposalAttributes.tsx
```diff
@@ -95,6 +96,20 @@ export const ProposalAttributes: React.FC<ProposalAttributesProps> = ({
...
 		if (isFromSnapshotWithMultipleChoices(proposal)) {
 			parsedAttributes = parsedAttributes.slice(1);
+		} else {
+			parsedAttributes.splice(2, 0, { label: 'Type', value: 'Funding' });
+			parsedAttributes.splice(3, 0, {
+				label: 'Amount',
+				value:
+					formatTotalAmountOfTokenMetadata(proposal.metadata)?.toString() ||
+					'-',
+			});
```
colbr, line 106: nit: is there a reason we're using `splice` here instead of just `parsedAttributes[2] = { ... }` or `parsedAttributes.push`?

colbr (changes requested):
UI is looking good, but a few important code changes to work through.

The main thing I'm noticing is that your hooks are doing _a lot_, and it makes the logic hard to follow. The components _look_ clean, but they're actually very hard to follow because to work out how they work, you have to constantly cross-reference the hook controlling them. 

For example, `useCreateProposal` is handling UI state for showing/hiding the "Connect Wallet" modal. This would be much cleaner if this state was just in the component which could render the modal. I don't think there are any benefits to handling this UI state in a hook, especially not when it comes at the cost of DX/readability. I don't actually think this hook needs to exist at all as it's (mostly) all just UI state.

Similar thing for `useCreateProposalForm`. It should responsible for handling form data, and *maybe* keeping track of where we are in the process of submitting (i.e. `is submitting`, `error`, `success`). Instead, it's handling that as well as navigation and UI state for the confirmation modal.

I also noticed some things that weren't picked up in the design review, I'll chuck them in the Notion card.

PR #586, opened 2022-05-25
feat: Moto Claim Flow (DomW)

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/dApp-Team-Dash-78b824c0cd0e46c59c19dfe103ae929a?p=73df6ef6699c44319d8990771ec1134d)

-----

## 1. Pull request checklist

- [x] Notion card has been moved to the Code Review column
- [x] Notion card has a link to this PR
- [x] A reviewer has been assigned to the Notion card

## 2. PR type
- Feature

## 3. What is the old behaviour?
Claim flow does not exist
Unable to open claim nft via Wheels NFT View

## 4. What is the new behaviour?
Claim flow added
Able to open claim nft via Wheels NFT View

PR #609, opened 2022-07-04
feat: moto claim tweaks

domw30:
Extended PR from https://github.com/zer0-os/zNS-dapp/pull/586

This branch and PR has been created from feat/zauction-v2.1 which includes the above PR changes

### DIFF
--- a/src/components/MintPreview/MintPreview.tsx
+++ b/src/components/MintPreview/MintPreview.tsx
@@ -29,6 +29,8 @@ import {
 	BUTTON_TEXT,
 } from './MintPreview.constants';
 import { ZNA } from 'constants/zna';
+import { ROUTES } from 'constants/routes';
+import { IS_DEFAULT_NETWORK } from 'constants/domains';
 
 //- Assets Imports
 import questionMark from './assets/question-mark-icon.svg';
@@ -54,7 +56,9 @@ const MintPreview = (props: MintPreviewProps) => {
 		statusText?: string,
 	) => {
 		const link = zNAToLink(nft.zNA);
-		const parsedLink = getNetworkZNA(link);
+		const parsedLink = IS_DEFAULT_NETWORK
+			? getNetworkZNA(link)
+			: ROUTES.MARKET + '/' + getNetworkZNA(link);
 
 		// Check for Claim domain name
 		const claimDomainName = nft.zNA === ZNA.CLAIM_NFT_DOMAIN_ROOT;
@@ -87,11 +91,8 @@ const MintPreview = (props: MintPreviewProps) => {
 							<div className={styles.InfoSection}>
 								<h3>{nft.title}</h3>
 
-								<Link className={styles.Link} to={link}>
-									{/* // TODO: Need to check if this is working as expected - resolved as merge conflict */}
-									{/* 0://{truncateDomain(nft.zNA, MAX_CHARACTER_VALUE)} */}
+								<Link className={styles.Link} to={parsedLink}>
 									{truncateDomain(nft.zNA, MAX_CHARACTER_VALUE)}
-
 									{claimDomainName && '.?'}
 								</Link>
 							</div>
--- a/src/components/MintPreview/MintPreview.tsx
+++ b/src/components/MintPreview/MintPreview.tsx
@@ -122,11 +122,11 @@ const MintPreview = (props: MintPreviewProps) => {
 
 							{nft.stakeAmount && (
 								<div>
-									{nft.stakeAmount && nft.stakeAmount.length > 0 ? (
+									{nft.stakeAmount.length > 0 && (
 										<p style={{ marginTop: '16px' }}>
 											Stake Amount: {nft.stakeAmount} LOOT
 										</p>
-									) : null}
+									)}
 									<div style={statusStyle}>{statusText}</div>
 								</div>
 							)}
--- a/src/components/MintPreview/MintPreview.tsx
+++ b/src/components/MintPreview/MintPreview.tsx
@@ -30,7 +30,6 @@ import {
 } from './MintPreview.constants';
 import { ZNA } from 'constants/zna';
 import { ROUTES } from 'constants/routes';
-import { IS_DEFAULT_NETWORK } from 'constants/domains';
 
 //- Assets Imports
 import questionMark from './assets/question-mark-icon.svg';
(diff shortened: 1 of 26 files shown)


src/lib/hooks/useOwnedDomains.ts
```diff
@@ -38,10 +38,13 @@ const useOwnedDomains = (
 			const owned = await sdk.getDomainsByOwner(account);
 			// TODO: Optimize this
 			const domainsPaymentTokenData = owned.map(async ({ id }) => {
-				const paymentToken = await sdk.zauction.getPaymentTokenForDomain(id);
+				// const paymentToken = await sdk.zauction.getPaymentTokenForDomain(id);
 				const paymentTokenInfo: PaymentTokenInfo = {
-					...(await sdk.zauction.getPaymentTokenInfo(paymentToken)),
-					id: paymentToken,
+					...(await sdk.zauction.getPaymentTokenInfo(
+						// TODO :: REPLACE WITH paymentToken
```
domw30, line 44: This is required to allow QA to test all claimed motos appearing in profile issue.

src/lib/utils/searchResults.ts
```diff
@@ -0,0 +1,17 @@
+//- Library Improts
+import { DisplayParentDomain } from 'lib/types';
+
+export const filterSearchResultsByNetwork = (
+	results: DisplayParentDomain[],
```
padala, line 5: @domw30  What is the purpose of this util?

domw30: @padala, the intention was to filter domains from another network when ```REACT_APP_NETWORK=wilder```. Say for example the network is set to ```tester```, when searching the results will show domains on ```wilder``` network. If selected, you'd hit a page not found. I thought now we switch between networks this may be useful? Let me know your thoughts and feedback, much appreciated Subbu thanks.

colbr: @domw30 could work out nicer to append the root network to any search queries if there is one - that way we don't have to do any additional filtering in the UI.

domw30: @colbr, do you mean append the root network as it currently is? 

Screenshot 2022-07-05 at 17 24 30

colbr: Hmm, on second thought, doing it your way makes more sense. Appending the root network to the search query would prevent people from searching "wilder.beasts" (as the query would be "wilder.wilder.beasts"). I'll review the rest of this one today, just saw this comment as I was scrolling and thought I'd chime in!

src/components/MintPreview/MintPreview.tsx
```diff
@@ -54,7 +56,9 @@ const MintPreview = (props: MintPreviewProps) => {
 		statusText?: string,
 	) => {
 		const link = zNAToLink(nft.zNA);
-		const parsedLink = getNetworkZNA(link);
+		const parsedLink = IS_DEFAULT_NETWORK
+			? getNetworkZNA(link)
+			: ROUTES.MARKET + '/' + getNetworkZNA(link);
```
colbr, line 61: Just checking -

I noticed in TransferPreview we're appending `ROUTES.MARKET` to the network ZNA regardless of whether or not `IS_DEFAULT_NETWORK === true`. Should we also be doing that here, or is there an edge case I'm missing?

src/components/ZNALink/ZNALink.tsx
```diff
@@ -24,7 +24,7 @@ const ZNALink: React.FC<ZNAProps> = ({ className, style }) => {
 	const zna = zNAFromPathname(pathname);
 	const app = appFromPathname(pathname) + '/';
 	const isRootDomain = zna.length === 0;
-	const isRootWithSubDomain = zna.split('.').length > 2;
+	const isRootWithSubDomain = zna.split('.').length >= 2;
```
colbr, line 27: True - nice find.

src/lib/utils/searchResults.ts
```diff
@@ -0,0 +1,17 @@
...
+	if (!results) {
+		return results;
+	} else if ((process.env.REACT_APP_NETWORK ?? '') === '') {
+		return results;
+	} else {
+		const onNetworkSearchResults = results?.filter(
+			(result) => result.name.split('.')[0] === process.env.REACT_APP_NETWORK,
+		);
+		return onNetworkSearchResults;
+	}
```
colbr, line 16: nit: could be cleaned up a little
```js
if(results && DEFAULT_NETWORK) {
    return results.filter((r) => r.name.startsWith(DEFAULT_NETWORK + '.'))
}
return results
```
_(not entirely sure if this logic is accurate or the var names are accurate)_

If you have logic statements which return the same value (in this case `results`) it's generally easier to follow if you combine them. I also like to use an early return instead of `else` when you're working with relatively simple logic and no `else if` - but that's just preference!

colbr (approved):
A question and a nit - but I'll give it an LGTM and fire emoji 🔥

src/lib/utils/searchResults.ts
domw30: Thanks mate! 🚀

src/components/MintPreview/MintPreview.tsx
domw30: Hmm yeah I just looped back on this one. I did this because ```const link = zNAToLink(nft.zNA);``` being set just above line 59 here, appends '/market/` route in the return. Thought ```zNAToLink``` may have been used elsewhere but its not, so I've removed the '/market/' from that util and made the change here to mirror the same as ```TransferPreview``` variable ```parsedLink```

Testing with no network env set:
Screenshot 2022-07-06 at 12 03 06


Testing with wilder network env set:
Screenshot 2022-07-06 at 12 09 42

Nice! Thanks for pointing that out @colbr

src/lib/utils/searchResults.ts
domw30: Yeah that makes sense, noticed similar comments in the other open PR, will take this on board :) Thanks! Updated

src/containers/flows/ClaimNFT/components/WizardSteps/Claiming/Claiming.tsx
```diff
@@ -55,18 +55,12 @@ const Claiming = ({
...
 
 	const validQuantity =
-		Number(quantity) <= totalEligibleDomains &&
+		Number.isInteger(Number(quantity)) &&
 		Number(quantity) > 0 &&
-		Number(quantity) <= maxQuantityLimit &&
-		quantity !== undefined &&
-		!isNaN(Number(quantity)) &&
-		Number(quantity) % 1 === 0;
+		Number(quantity) <= Math.min(totalEligibleDomains, maxQuantityLimit);
```
colbr, line 63: Nice, looking cleaner 🧹

PR #619, opened 2022-07-12
fix(claim-check): improve checking UX

colbr:
## 1. PR type
 

Feature

## 2. What is the old behaviour?

The "is claimable" NFT check only allowed for hex addresses.

## 3. What is the new behaviour?

- "Is claimable" check now allows for integer token IDs
- Added an error message for when a user inputs an IPFS hash (common mistake)


domw30 (approved):
LGMT! nice !

domw30:
There's some console logging  in useIsClaimable.tsx but we can remove later

PR #626, opened 2022-07-19
fix: accept bid - nft view data refetch

domw30:
[Associated Notion Card](https://www.notion.so/zerotech/0-17-0-zAuction-2-1-network-switcher-a2c4a3409df541a197682d740babc572?p=15db0ce0bded410ab81fc65d3b64f972)

-----

## 1. Pull request checklist

## 2. PR type
- [x] bug

When accept bid flow is successful on NFT View page, the actions do not update to reflect the user no longer owns the domain. Also the history items do not update to reflect the sale.


src/containers/flows/AcceptBid/AcceptBid.tsx
```diff
@@ -43,12 +43,12 @@ type AcceptBidProps = {
 	assetUrl: string;
 	creatorId: string;
 	domainTitle: string;
-	domainId: string;
 	domainName: string;
 	walletAddress: string;
 	highestBid?: string;
 	onClose: () => void;
 	paymentTokenInfo: ConvertedTokenInfo;
+	setIsAcceptBidProcessing: (state: boolean) => void;
```
colbr, line 51: Could be beneficial to make this one more generic, perhaps it could be a callback which is fired for each step of the AcceptBid flow.

`onChangeStep: (step: Step) => void`

```js
const changeStep = (step: Step) => {
    onChangeStep(step);
    setCurrentStep(step);
}
...
const onAcceptBid = () => {
    changeStep(Step.Processing);
    try {
        await accept(acceptingBid!);
        changeStep(Step.Success);
    } catch (e) {
        changeStep(Step.Failed);
    }
}
```

Or something like that?

src/containers/lists/BidList/BidList.tsx
```diff
@@ -1,187 +1,120 @@
...
-										{Number(bid.expireBlock) <= blockNumber && (
-											<div>Expired</div>
-										)}
-									</>
-								)}
-							</li>
-						))}
-					</ul>
-				</aside>
+		<aside className={`${styles.Container} border-rounded border-primary`}>
```
colbr, line 58: nit: `classNames(...)`

src/containers/lists/BidList/BidList.tsx
```diff
@@ -1,187 +1,120 @@
...
+			<div className={styles.Header}>
+				<h4>All bids</h4>
+				<hr className="glow" />
+			</div>
+			{isAcceptBidProcessing && (
+				<p className={styles.Pending}>Accept bid transaction pending.</p>
 			)}
-		</>
+			<ul className={isAcceptBidProcessing ? styles.Accepting : ''}>
```
colbr, line 66: nit: use `classnames/bind`

src/containers/lists/BidList/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+					bids={bids}
+					paymentTokenInfo={paymentTokenInfo}
+					isAcceptBidEnabled
+					isAcceptBidProcessing={isAcceptBidProcessing}
+					openAcceptBid={handleAcceptBid}
+				/>
+			)}
+		</>
+	);
+};
```
colbr, line 92: nit: can make this a little bit cleaner by

```js
if(isAcceptBidModalOpen && onAcceptSuccess) {
    return <acceptbid ... />
} else {
    return <bidlist ... />
}
```

Just to save using a fragment and nesting logic in JSX.

colbr (commented):
Looking solid, nice one 🚀

src/containers/flows/AcceptBid/AcceptBid.tsx
domw30: This is being used to handle disabled UI on for bid list, a change in step wouldn't be required. So if a user clicks out of the accept bid modal while transaction is pending, they will see: 

Screenshot 2022-07-20 at 13 46 53

colbr: Gotcha makes sense!

PR #627, opened 2022-07-20
feat: improve loading efficiency of Subdomain Table

colbr:
- Memoise some components
- Combine state updates in CurrentDomainProvider which were previously split by async logic
- Re-enable stat cards


padala (approved):
@colbr  Looks good to me.

domw30 (approved):
looks good ! 🧹

PR #636, opened 2022-07-22
Fix/tests

domw30:
Fix failing tests for release and bump version 0.17.0


src/components/Dropdowns/OptionDropdown/index.ts
```diff
@@ -1 +1 @@
-export * from './OptionDropdown';
+export { default as OptionDropdown } from './OptionDropdown';
```
padala, line 1: @domw30  looks like build is failing.

PR #645, opened 2022-08-02
feat: handle delayed voting

colbr:
[Associated Notion Card](https://www.notion.so/zerotech/Handle-delayed-voting-bfaf57923c814128a86ca9b56ed762bd)

-----

## 1. Pull request checklist

## 2. PR type
 

Feature

Voting delays were not implemented - instead the UI would let users go through the voting process, which would end in an error.

- Shows a timer if the proposal voting hasn't started.
- Shows "Voting Starts" before voting has started, "Voting Started" after.
- (extra) Changed "Voting System" attribute so it's not hard-coded.


EthFish (approved):
Looks good!

PR #655, opened 2022-10-17
Goerli Migration

colbr:


src/lib/paymentToken.ts
```diff
@@ -11,12 +12,13 @@ const getPaymentTokenInfo = async (
...
 	tokenId: string,
 ): Promise<ConvertedTokenInfo> => {
+	const { rinkeby, goerli } = ZContracts.zer0ProtocolAddresses;
 	if (paymentTokenCache[tokenId]) {
 		return paymentTokenCache[tokenId];
 	} else {
 		if (
-			tokenId.toLowerCase() ===
-			'0x3Ae5d499cfb8FB645708CC6DA599C90e64b33A79'.toLowerCase()
+			tokenId.toLowerCase() === rinkeby?.tokens?.wildToken?.toLowerCase() ||
```
domw30, line 20: Is there a reason to keep Rinkeby related configurations?

domw30, line 20: Is there a reason to keep Rinkeby related configurations?

domw30 (approved):
Changes look good to me Subbu

src/lib/paymentToken.ts
padala: We can actually get rid of it because this never gets used. Will update it. Thanks

